PR #103, opened 2015-02-10
Update RSpec 3

mikeastock:
This PR upgrades RSpec from 2 to 3.


mikeastock:
@sudara @wioux So my build seems to be missing
```
export S3_CREDENTIALS=[secure]
```
Any help figuring out why it's not around would be greatly appreciated

sudara:
Awesome!! 

Are we sure ruby-mp3info still functions as normal? I think this may have been locked for some compat reason, not sure why now.

Are you talking about 

```  
def s3_credentials
   ENV['S3_CREDENTIALS'].split(':')
end
```

If so, this is used by wad — which is a fngtps thing! https://github.com/Fingertips/WAD — not sure this is relevant except for running on travis. Also, if travis supports dependency caching for open source / free projects now, we can toss wad (I know they support it for paid projects).

sudara:
Also, @wioux has a dependency branch started, you may wanna check with him. I'd love to update everything, but I'm not personally confident we have solid spec coverage on things like authentication, mp3 info extraction, paperclip, etc.

sudara:
Looks like deps caching isn't restricted to paid projcets anymore? Maybe we can toss WAD http://docs.travis-ci.com/user/caching/#Caching-directories-(Bundler%2C-dependencies)

PR #109, opened 2015-02-18
Add Bugsnag

mikeastock:
Before we can merge this we need to set ENV["BUGSNAG_API_KEY"] to our
API key in production

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,7 @@ gem "awesome_print", :require => 'ap'
 gem "aws-sdk"
 gem "paperclip"
 gem "ruby-mp3info", :require => 'mp3info'
-gem "mime-types" 
+gem "mime-types"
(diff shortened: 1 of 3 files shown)


sudara:
WOO bugs.

Is there a way to roll with config? I don't like "ENV vars as config" unless it's a rails server/server requirement.

sudara:
looks like we can add the key to the alonetone.yml on the server and do

```
Bugsnag.configure do |config|
  config.api_key = Alonetone.bugsnag_key
end```

In general, we should think about moving to the rails 4.2 way of handling yml config too: http://www.justinweiss.com/blog/2014/08/25/the-lesser-known-features-in-rails-4-dot-2/

sudara:
Merged, but going to use alonetone.yml and `config_for` on master.

PR #128, opened 2016-02-05
Greenfield playlist downloads

sudara:


sudara:
Going to merge this shortly, I can add track titles in the console for now.

Issue #144, opened 2017-06-24
Fix dupes of Recent Listeners


sudara:
Because of the upgrade to Mysql 5.7, this query stopped working. Need to get uniq/distinct listeners on the single track page.

mishraak:
I was trying to see if I can contribute to this project and hence checking if there were duplicates but for multiple tracks wasn't able to find any duplicates in the 'Recent Listeners'. Has this been fixed already? Am I missing something here?

sudara:
Hey @mishraak 

Thanks for wanting to help out!

I believe it's still an issue https://alonetone.com/gl1tchkids/tracks/psalty-as-fuck-1

This relationship needs to be unique:
https://github.com/sudara/alonetone/blob/master/app/models/asset.rb#L19-L21

PR #151, opened 2018-03-07
#144 fix duplicate recent listeners

mishraak:
"distinct" keyword to retrieve distinct users.


sudara:
Thanks for the Pull Request!

Let's get you up and running on the app — that will mean you can contribute much more quickly!

In this case, the line had to be:

`-> { distinct.order('listens.created_at DESC').limit(20) }`

Which actually caused an error since we updated to mysql 5.7. So I also had to set `sql_mode` for it to work.
https://github.com/sudara/alonetone/commit/6f1164baef7f0ad8d938d016e63343c16b1e79ad

Let me know if you want a slack login!

PR #181, opened 2018-08-15
Add Annotate gem and notes

innatewonderer:
I find it useful to be able to reference what attributes are still applicable to each model.

### DIFF
--- a/app/models/track.rb
+++ b/app/models/track.rb
@@ -1,3 +1,38 @@
+class Track < ActiveRecord::Base
+  belongs_to :playlist, :counter_cache => true, :touch => true
+  belongs_to :asset
+  belongs_to :user
+
+  scope :recent, -> { order('tracks.created_at DESC') }
+  scope :favorites, -> { where(:is_favorite => true).recent }
+  scope :favorites_for_home, -> { favorites.limit(5).includes({:user => :pic}, {:asset => {:user => :pic}}) }
+
+  delegate :length, :name, :to => :asset
+  acts_as_list :scope => :playlist_id, :order => :position
+
+  before_validation :ensure_playlist_if_favorite
+  validates_presence_of :asset_id, :playlist_id, :user_id
+
+  def asset_length
+    asset ? asset[:length] : 0
+  end
+
+  def self.most_favorited(limit=10, offset=0)
+    self.count(:all,
+      :conditions => ['tracks.is_favorite = ?', true],
+      :group      => 'asset_id',
+      :order      => 'count_all DESC',
+      :limit      => limit,
+      :offset     => offset
+    ).collect(&:first)
+  end
+
+  def ensure_playlist_if_favorite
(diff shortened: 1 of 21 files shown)


sudara:
How would you feel about putting the annotation at the bottom of the files for reference? I kind of enjoy being free from the clutter when I first open a file. I could maybe get used to it again if you prefer them up top...

innatewonderer:
I have no preference, as long as I can see the model's attributes. Let me see if there is a setting somewhere.

sudara:
Ok cool! One of the reason I originally removed them is they would get out of sync with the schema. Since I work on many apps where this is the case, I essentially just refer to the schema.rb to assess the attributes. This will be nice at the bottom if we can keep it up to date!

innatewonderer:
This should help I think:
```
To automatically annotate every time you run db:migrate, either run rails g annotate:install or add Annotate.load_tasks to your `Rakefile`. See the configuration in Rails section for more info.
```

sudara:
The rake task sounds like a great way to solve it!

config/environments/development.rb
```diff
@@ -1,7 +1,7 @@
 Rails.application.configure do
   # Verifies that versions and hashed value of the package contents in the project's package.json
-    config.webpacker.check_yarn_integrity = true
-    config.webpacker.check_yarn_integrity = true  # Settings specified here will take precedence over those in config/application.rb.
+    config.webpacker.check_yarn_integrity = false
+    config.webpacker.check_yarn_integrity = false  # Settings specified here will take precedence over those in config/application.rb.
```
sudara: Are these changes related? I think we want this to be set back to `true` so that yarn complains when package.json changes. Not sure why there were two lines though, maybe we can delete one?

app/models/track.rb
```diff
@@ -1,14 +1,22 @@
...
+#  playlist_id :integer
+#  user_id     :integer
+#
+# Indexes
+#
+#  index_tracks_on_asset_id     (asset_id)
+#  index_tracks_on_is_favorite  (is_favorite)
+#  index_tracks_on_playlist_id  (playlist_id)
+#  index_tracks_on_user_id      (user_id)
 #
```
sudara: Hah, like this file had very very old leftovers in the header, would you mind moving it to the footer?

lib/tasks/auto_annotate_models.rake
```diff
@@ -0,0 +1,54 @@
...
+      'force'                     => 'false',
+      'classified_sort'           => 'true',
+      'trace'                     => 'false',
+      'wrapper_open'              => nil,
+      'wrapper_close'             => nil,
+      'with_comment'              => true
+    )
+  end
+
+  Annotate.load_tasks
```
innatewonderer: This should run annotate automagically on `rake db:migrate`. I'll test to confirm

innatewonderer:
@sudara mergeable

Issue #177, opened 2018-08-15
Get rid of old hash syntax 


sudara:
Lots of hashrockets laying around, including some mixing, which is kind of crazy. Probably time to up the hygiene:

`<%= link_to radio_path(:per_page => 40, :source => 'those_you_follow'), :title => 'alonetone radio: those you follow' do %>`

innatewonderer:
Shold we close this issue @sudara ?

PR #183, opened 2018-08-17
[WIP] Upgrading hash syntax in views

innatewonderer:
Rubocop does not want to go through views, so we'll do it manually.
Working on https://github.com/sudara/alonetone/issues/177


innatewonderer:
The only thing was that when I looked at their compatibility table 

| Version | branch | ruby | activerecord |
|-----------|-----------|-----------|-----------|
| 3 | 3-stable | >= 1.9.3 | >= 3.2, < 5.2 |

I was concerned that our `3.8` gem version was barely keeping up with `5.3` ActiveRecord dependancy. I checked locally and it seemed to have worked. No objections to keep it as is.

PR #188, opened 2018-08-22
Icons

sudara:


app/helpers/application_helper.rb
```diff
@@ -177,5 +177,9 @@ def sanitize_feed_content(html, sanitize_tables = false)
     sanitized
   end
 
+  def svg_path(svg)
+    File.join(Rails.root, 'app', 'assets', 'images', svg)
+  end
+
```
sudara: @ofsound if you know where the svgs will live I can add that path here so you can just do `<%= render file: svg_path('play-icon.svg') %>` instead of <%= render file: svg_path('default/play-icon.svg') %>

PR #191, opened 2018-08-24
Add title to NewTrackUploaded email. Made some changes to text as well.

innatewonderer:


app/views/asset_notification/upload_notification.text.erb
```diff
@@ -1,26 +1,26 @@
...
 
-Listen to it at alonetone:
+To listen to <%= @title %> on alonetone:
 <%= @play_link %>
 
-Download it directly:
+Or download it directly:
 <%= @download_link%>
 
------8<-------------------------
```
innatewonderer: I removed this... Not sure if it was intentional?

app/views/asset_notification/upload_notification.text.erb
```diff
@@ -1,26 +1,26 @@
...
 <% else %>
-<%= @name %> just uploaded a new track on alonetone. <%= @exclamation %>!
+New track by <%= @name.capitalize %> was just uploaded to alonetone! <%= @exclamation %>!
 
-It's so piping hot we don't even know the juicy details yet. <%= @name %> is still typing them in.
+It's so piping hot we don't even know the juicy details yet. <%= @name.capitalize %> is still typing them in.
 <% end %>
 
-Listen to it at alonetone:
+To listen to <%= @title %> on alonetone:
```
innatewonderer: Didn't want to break the `flow` of the email by sticking the title somewhere up top. Would this place be enough you think? Or we can do something more explicit in the first/second sentence.

app/mailers/asset_notification.rb
```diff
@@ -6,6 +6,7 @@ def upload_notification(asset, email, _sent_at = Time.now)
     @description = asset.description
     @name = asset.user.name
     @user = asset.user
+    @title = asset.title.present? ? asset.title : "new track"
```
innatewonderer: I just tested uploading a new file, and it seems to grab `filename` and save it as `title` by default. Should I just leave this like this?  I wanted to add a placeholder in case if title is not available in some older songs... going to unpack the db dump and check actually :)

PR #198, opened 2018-09-04
Add `delete user` functionality

innatewonderer:
- [x]  As an admin I should be able to delete user
- [x] Make sure `delete my account` is fixed for user (white=false theme)


innatewonderer:
Too many conflicts to fix with master. will create a new PR

PR #241, opened 2018-10-05
Convert settings to store

matthewheath:
Convert User profile settings to use ActiveRecord::Store and update UsersController params to permit settings parameters.


sudara:
Thanks so much!!

Issue #205, opened 2018-09-22
Don't allow user comments on their own track to show up on home page


sudara:
This is often used as a way for folks to get more attention...

<img width="597" alt="image" src="https://user-images.githubusercontent.com/472/45919113-2c71f000-be91-11e8-80c7-80236806befb.png">

PR #244, opened 2018-10-08
User comments

innatewonderer:
Fixes https://github.com/sudara/alonetone/issues/205

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -16,7 +16,7 @@ class Asset < ActiveRecord::Base
   has_many :comments, as: :commentable, dependent: :destroy
 
   has_many :listeners,
-    -> { distinct.order('listens.created_at DESC').limit(20) },
+    -> { select("distinct listens.*").order('listens.created_at DESC').limit(20) },
     through: :listens
 
   has_many :favoriters,
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -287,7 +287,7 @@ class AssetsController < ApplicationController
     @favorites = Track.favorites_for_home
     @popular = Asset.published.order('hotness DESC').includes(user: :pic).limit(5)
     @playlists = white_theme_enabled? ? Playlist.for_home.limit(4) : Playlist.for_home.limit(5)
-    @comments = admin? ? Comment.last_5_private : Comment.last_5_public
+    @comments = admin? ? Comment.last_5_private : Comment.to_other_members.last_5_public
     @followee_tracks = current_user.new_tracks_from_followees(5) if user_has_tracks_from_followees?
   end
 
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -9,6 +9,7 @@ class Comment < ActiveRecord::Base
   scope :last_5_private,     -> { on_track.include_private.limit(5).preload(commenter: :pic, commentable: { user: :pic }) }
   scope :last_5_public,      -> { on_track.only_public.limit(5).preload(commenter: :pic, commentable: { user: :pic }) }
   scope :made_between,       ->(start, finish) { where('comments.created_at BETWEEN ? AND ?', start, finish) }
+  scope :to_other_members,   -> { where("commenter_id != user_id") }
 
   has_many :replies, as: :commentable, class_name: 'Comment'
 


innatewonderer:
@sudara What do you have in mind to help with `comment index issues`? 
I see that user_id isn't indexed, we can add that... Do you have something more specific in mind?

sudara:
@innatewonderer Comments#index has been one of the slower actions, in part because the queries seem to be slow (also a lot of n+1s in the views, it seems) https://oss.skylight.io/app/applications/QMmsxBDrac9Q/recent/6h/endpoints/CommentsController%23index?responseType=html

The existing indexes are:

```
    t.index ["commentable_id"], name: "index_comments_on_commentable_id"
    t.index ["commenter_id"], name: "index_comments_on_commenter_id"
    t.index ["created_at"], name: "index_comments_on_created_at"
    t.index ["is_spam"], name: "index_comments_on_is_spam"
```

Right now two queries tend to fire, one is a count for will_paginate:

image

And the other is either a per-user or general query to grab the comments:

image

Both of these together take over 300ms typically, sometimes the per user one takes 200ms. They should probably take more like 20ms.

Comments are polymorphic in the database, they have a `commentable_type` — this is because historically, one could comment on `updates` (blog entries) and `features` as well as `assets`— neither of these exist anymore. We should think about removing the polymorphism.

For now, we could adjust the index to include the polymorphic type as described here: https://mauricio.github.io/2008/09/27/handling-database-indexes-for-rails-polymorphic-associations.html

I don't think this PR will do much detriment the score much, but because our indexes already suck (the existing indexes only helps the db reduce the table down to 57k rows currently) then I thought it would be a good time to evaluate.

Here's the explain on a typical comments query:

```
irb(main):010:0> User.first.comments.on_track.public_or_private(true).page(1).explain
=> EXPLAIN for: SELECT  `comments`.* FROM `comments` WHERE `comments`.`user_id` = 1 AND `comments`.`commentable_type` = 'Asset' AND `comments`.`is_spam` = FALSE ORDER BY comments.id DESC, id DESC LIMIT 30 OFFSET 0
+----+-------------+----------+------------+------+---------------------------+---------------------------+---------+-------+-------+----------+-------------+
| id | select_type | table    | partitions | type | possible_keys             | key                       | key_len | ref   | rows  | filtered | Extra       |
+----+-------------+----------+------------+------+---------------------------+---------------------------+---------+-------+-------+----------+-------------+
|  1 | SIMPLE      | comments | NULL       | ref  | index_comments_on_is_spam | index_comments_on_is_spam | 2       | const | 57392 |      1.0 | Using where |
+----+-------------+----------+------------+------+---------------------------+---------------------------+---------+-------+-------+----------+-------------+
1 row in set (0.00 sec)
```

Adding the extra condition in this PR just adds a bit of work at the end:

```
irb(main):011:0> User.first.comments.on_track.public_or_private(true).where("commenter_id != user_id").page(1).explain
=> EXPLAIN for: SELECT  `comments`.* FROM `comments` WHERE `comments`.`user_id` = 1 AND `comments`.`commentable_type` = 'Asset' AND `comments`.`is_spam` = FALSE AND (commenter_id != user_id) ORDER BY comments.id DESC, id DESC LIMIT 30 OFFSET 0
+----+-------------+----------+------------+------+----------------------------------------------------------+---------------------------+---------+-------+-------+----------+-------------+
| id | select_type | table    | partitions | type | possible_keys                                            | key                       | key_len | ref   | rows  | filtered | Extra       |
+----+-------------+----------+------------+------+----------------------------------------------------------+---------------------------+---------+-------+-------+----------+-------------+
|  1 | SIMPLE      | comments | NULL       | ref  | index_comments_on_commenter_id,index_comments_on_is_spam | index_comments_on_is_spam | 2       | const | 57392 |      0.5 | Using where |
+----+-------------+----------+------------+------+----------------------------------------------------------+---------------------------+---------+-------+-------+----------+-------------+
1 row in set (0.00 sec)
```

For now, I added an index for type/spam/private and all should be better, the indexes help the db whittle things down to 60 rows, and only then do other conditions get evaluated:

```
irb(main):002:0> User.first.comments.on_track.public_or_private(true).where("commenter_id != user_id").page(1).explain
  User Load (0.5ms)  SELECT  `users`.* FROM `users` ORDER BY `users`.`id` ASC LIMIT 1
  Comment Load (19.9ms)  SELECT  `comments`.* FROM `comments` WHERE `comments`.`user_id` = 1 AND `comments`.`commentable_type` = 'Asset' AND `comments`.`is_spam` = FALSE AND (commenter_id != user_id) ORDER BY comments.id DESC, id DESC LIMIT 30 OFFSET 0
=> EXPLAIN for: SELECT  `comments`.* FROM `comments` WHERE `comments`.`user_id` = 1 AND `comments`.`commentable_type` = 'Asset' AND `comments`.`is_spam` = FALSE AND (commenter_id != user_id) ORDER BY comments.id DESC, id DESC LIMIT 30 OFFSET 0
+----+-------------+----------+------------+-------+-------------------------------------------------------------------------------------------+---------+---------+------+------+----------+-------------+
| id | select_type | table    | partitions | type  | possible_keys                                                                             | key     | key_len | ref  | rows | filtered | Extra       |
+----+-------------+----------+------------+-------+-------------------------------------------------------------------------------------------+---------+---------+------+------+----------+-------------+
|  1 | SIMPLE      | comments | NULL       | index | index_comments_on_commenter_id,index_comments_on_commentable_type_and_is_spam_and_private | PRIMARY | 4       | NULL |   60 |      2.5 | Using where |
+----+-------------+----------+------------+-------+-------------------------------------------------------------------------------------------+---------+---------+------+------+----------+-------------+
1 row in set (0.00 sec)
```

sudara:
Added new index and removed unused indexes:

https://github.com/sudara/alonetone/commit/42d440f048c96dfdaccdd2604b28862ab9554d90

app/models/asset.rb
```diff
@@ -16,7 +16,7 @@ class Asset < ActiveRecord::Base
   has_many :comments, as: :commentable, dependent: :destroy
 
   has_many :listeners,
-    -> { distinct.order('listens.created_at DESC').limit(20) },
+    -> { select("distinct listens.*").order('listens.created_at DESC').limit(20) },
```
sudara: FYI, this broke production. Not sure why it was added, or if it was intentional... but it resulted in getting back fake users like this:

```
=> #<ActiveRecord::Associations::CollectionProxy [#<User id: 7599463, created_at: "2018-09-29 22:16:11", updated_at: "2018-09-29 22:16:11", country: nil, city: nil>, #<User id: 7599462, created_at: "2018-09-29 21:15:19", updated_at: "2018-09-29 21:15:19", country: nil, city: nil>]>
```

innatewonderer: Weird... `distinct.order('')` was breaking locally... maybe version discrepancy? but i doubt that 🤔

Issue #269, opened 2018-10-29
[Admin] setup `/assets` namespace


innatewonderer:
- [ ] setup assets admin area
- [ ] feedback from [Sudara](https://github.com/sudara/alonetone/pull/254#issuecomment-433312721)

The assets table would look similarly:

timestamp | IP/login | Asset ID (view) | Asset Title | Truncated asset description... | [Mark as Spam] | [Mark All from User/IP as Spam] | [Delete User (w/confirmation)] 
---|---|---|---|---|---|---|---|

Issue #268, opened 2018-10-29
[Admin] setup `/user` namespace


innatewonderer:
- [ ] setup `/users` admin section
- [ ] incorporate feedback from [Sudara](https://github.com/sudara/alonetone/pull/254#issuecomment-433312721) to get us started

timestamp | IP | user login | user bio truncated... |  [Mark All Tracks As Spam] | [Add IP to Blocklist] | [ Delete User] 
---|---|---|---|---|---|---|

sudara:
I'm thinking we want a "filter by" element that breaks down user types and provides counts that includes the categories here https://alonetone.com/users but with some more options for admin:

Trackless Unactivated older than 4 months (443)
```
User.where('perishable_token is not null').where('created_at < ?', 4.weeks.ago).where('assets_count = 0')
```

Musicians who reset pass and never clicked (33)
```
User.where('perishable_token is not null').where('assets_count > 0')
```

Trackless never updated (16085)
```
User.where('DATE(created_at) = DATE(updated_at)').where('assets_count = 0')
```

Musicians using bandwidth this month (447)
```
User.order('bandwidth_used desc').where('created_at > ?', 4.weeks.ago).limit(25)
```

Musicians using most bandwidth total (25)
```
User.order('bandwidth_used desc').limit(25)
```

Issue #267, opened 2018-10-29
[Admin] Setup `/comments` namespace


innatewonderer:
- [x] Setup `/comments` namespace
- [ ] Incorporporate feedback from [Sudara](https://github.com/sudara/alonetone/pull/254#issuecomment-433312721)
- [x] Try out `pagy` from https://github.com/sudara/alonetone/issues/247

For example, the comments section could start off looking like this, with actions performed as ajax inline:

timestamp | IP/login | Asset ID (view) | Truncated text of comment.... | [Mark as Spam] | [Mark All from User/IP as Spam] | [Add <ip> to blocklist]
---|---|---|---|---|---|---|

Issue #266, opened 2018-10-29
[Admin] Setup main /admin space


innatewonderer:
Setup `/admin` namespace, routes and permissions.

PR #278, opened 2018-11-05
Admin Comments section

innatewonderer:
- [x] Setup pagy gem for pagination https://github.com/sudara/alonetone/issues/247
- [x] setup basic admin namespace https://github.com/sudara/alonetone/issues/266 
- [x] setup admin/comments routes/controller/view and specs https://github.com/sudara/alonetone/issues/267
- [ ] setup admin/user routes/controller/view and specs https://github.com/sudara/alonetone/issues/268
- [ ] setup admin/asset routes/controller/view and specs https://github.com/sudara/alonetone/issues/269


app/models/comment.rb
```diff
@@ -94,6 +94,10 @@ def is_deliverable?
   def truncate_user_agent
     self.user_agent = user_agent.try(:slice, 0, 255)
   end
+
+  def body_snippet
+    self.body.truncate(100)
```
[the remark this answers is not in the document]
sudara: `self` is only needed on assignment — but also, we probably want to use Rails' view helper https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-truncate

app/controllers/admin/comments/mark_as_spam_controller.rb
```diff
@@ -0,0 +1,17 @@
+class Admin::Comments::MarkAsSpamController < Admin::BaseController
```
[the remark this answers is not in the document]
sudara: This bit us on another project recently, basically Rails' can't code load properly without explicit nesting because it might not have any idea what `Comments` is, a module or class. More here https://edgeguides.rubyonrails.org/autoloading_and_reloading_constants.html#nesting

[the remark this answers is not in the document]
sudara: Oh, sorry, just saw this was `Outdated` so I'll dismiss.

Gemfile
```diff
@@ -69,6 +69,10 @@ gem 'skylight'
 gem 'sidekiq'
 gem 'dalli'
 
+# admin functionality
+gem "datagrid"
```
sudara: Can we remove `datagrid` until we identify an explicit need on the frontend? I prefer to be very pessimistic on dependencies when possible.

app/controllers/admin/comments/add_to_blocklist_controller.rb
```diff
@@ -0,0 +1,12 @@
+module Admin
+  module Comments
+    class AddToBlocklistController < Admin::BaseController
```
sudara: Couple thoughts, you can hit me up in slack to hash it out:

1. I think we'll need to add/remove/index the blocklist (so the controller should be probably named `BlocklistController`)

2. It's IPs that we are managing (not comments) and we'll also be blocking them based on tracks, users, etc — so I'm thinking we probably want it to be `Admin::BlocklistController`

app/controllers/admin/comments/mark_as_spam_controller.rb
```diff
@@ -0,0 +1,21 @@
+module Admin
+  module Comments
+    class MarkAsSpamController < Admin::BaseController
```
sudara: What do you think about having this just be in `Admin::CommentsController`? 

I guess the question is if we are doing something RESTy or not. If so, we'd want to name this `Admin::Comments::SpamController` (since we'll be doing more than marking as spam, we'll be unspamming things too). But I'm leaning towards thinking we won't have anything REST-like here, I can only think of these actions right now:

1. Listing comments
2. Marking one as spam
3. Marking one as ham
4. Marking all from an ip as spam

app/controllers/admin/comments_controller.rb
```diff
@@ -0,0 +1,7 @@
+module Admin
+  class CommentsController < Admin::BaseController
+    def index
+      @pagy, @comments = pagy(Comment.all)
```
sudara: Woo pagy!

app/models/comment.rb
```diff
@@ -90,6 +90,10 @@ def is_deliverable?
   def truncate_user_agent
     self.user_agent = user_agent.try(:slice, 0, 255)
   end
+
+  def body_snippet
+    body.truncate(100)
```
sudara: It's probably a bit nicer to use Rails' truncate stuff as a view helper: https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-truncate

app/controllers/admin/comments/mark_as_spam_controller.rb
sudara: Hmm, and then I guess I'm dreaming of a button that says "Delete All Spam older than 30 days"

Gemfile
innatewonderer: Oh yep, Leftover from when I was experimenting with using it.

app/controllers/admin/comments/mark_as_spam_controller.rb
innatewonderer: Yes, I've been trying to be more RESTful. I think this way will force us to keep files small and specific... but OK to go either way honestly.

app/controllers/admin/comments_controller.rb
innatewonderer: I'm going to do some research on how to actually test performance for this :) Shouldn't be too complicated to switch other controllers/views to this just time and 👀 .

app/controllers/admin/comments/mark_as_spam_controller.rb
innatewonderer: Also, @sudara ... `Marking one as ham` ... you telling me `ham` is not a typo? :)

sudara: Thinking about the REST thing.... In order for REST to make sense, we'd need identify the entity/element we are operating on in collection/singular. `MarkAsSpam` doesn't work for me as entity we are listing/creating/destroying; it's not a noun that we are going to perform actions on.

`Comment` could be the element (`Admin::CommentsController`) and we could use member/collection routes to add actions that adjust the comment's attributes https://guides.rubyonrails.org/routing.html#adding-more-restful-actions — this is my preferred route because "spam" isn't an entity, it's an attribute that we are modifying, so it makes intuitive sense.

It could also be `Admin:Comments::Spam` if you want to view `Spam` as the noun being created/destroyed. I'd argue this would end up less clear to understand and read. What would marking something as "not spam" be? Using `DELETE` for that would be like  `destroying a comment spam` which is hard to understand I think.

But maybe there's a noun that would work? In this blog post, they talk about going more granular than the data model behind the API, one of the examples is `ChangeOfAddress`: https://www.thoughtworks.com/insights/blog/rest-api-design-resource-modeling

Hmmm, something like `MarkedSpam` might work? (a bit clunky). 

The action being performed as 1) informing an external API that they mis-categorized a comment (as "spam" or "not spam") and 2) changing the visibility of the comment on the site. I dunno, to me these keep seeming like `PUT` actions on a `Comment` resource, but maybe you can find a noun that works?

Bottom line is I'm always fine with going more granular given that we "list/create/update/destroy" a entity/resource/noun (otherwise we aren't actually doing REST)

> you telling me ham is not a typo? :)

Hehe, yeah, I actually don't like "ham" as the opposite of spam, it's confusing, but I guess it's a thing? 🐷 `mark_as_spam` and `mark_as_not_spam` sound better 🍡

app/controllers/admin/comments_controller.rb
sudara: OMG, I'm too excited. I'd be curious just to see what kinds of requests it makes (as seen in logs)

app/models/comment.rb
sudara: My gut says since the truncate helper is there, let's use it. I don't intuitively understand why we'd want to add code/abstraction to accomplish the same end result, but take that with a grain of salt and feel free to explain! I'm sorta a vanilla-rails person who prioritizes claritty and straightforwardness and gets confused easily with patterns and so on. My day job is basically unravelling patterns people add to their rails apps to regain performance/clarity :D I do think there are a few of places in alonetone in need of encapsulation (for example user names / profile pictures)...

app/controllers/admin/comments/mark_as_spam_controller.rb
sudara: Oooh, maybe `SpamReport` or `SpamStatus`? Still might be not really REST-ful because we'd actually be listing comments, not SpamReports, and we'd never update or destroy a SpamReport.... hm...

app/controllers/admin/comments_controller.rb
innatewonderer: Well, for starters :) we'd go from this:
```
  <div id="left" class="comments_area box">
    <h1>Recent Comments</h1>
    <div class="mini_paginator">
      <%= will_paginate @comments, param_name: :page%>
    </div>
    <%= render partial: 'shared/comment', collection: @comments %>
    <div class="mini_paginator">
      <%= will_paginate @comments, param_name: :page%>
    </div>
  </div>
```
to this:
```
  <div id="left" class="comments_area box">
    <h1>Recent Comments With Pagy</h1>
    <%== pagy_nav @pagy %>
    <%= render partial: 'shared/comment', collection: @comments_for_pagy %>
    <%== pagy_nav @pagy %>
  </div>
```

innatewonderer: Don't quote me on that, but here is what I did:
- setup some caching in my `development.rb` (it seems like pagy's benefits increase quite a lot once caching is enabled)
- performed simple Rails view benchmarking test 
Hard reloads: 
```
[22:35:16] (pagy-test) alonetone
$ tail -f log/development.log | grep 'Comments'
Processing by CommentsController#index as HTML
Comments (4156.4ms)
Comments with pagy (141.5ms)
Processing by CommentsController#index as HTML
Comments (1271.1ms)
Comments with pagy (111.4ms)
Processing by CommentsController#index as HTML
Comments (3525.9ms)
Comments with pagy (1504.4ms)
Processing by CommentsController#index as HTML
Comments (1607.2ms)
Comments with pagy (89.4ms)
Processing by CommentsController#index as HTML
Comments (376.9ms)
Comments with pagy (115.0ms)
```
Soft reloads
```
[22:36:52] (pagy-test) alonetone
$ tail -f log/development.log | grep 'Comments'
Processing by CommentsController#index as HTML
Comments (2595.4ms)
Comments with pagy (91.5ms)
Processing by CommentsController#index as HTML
Comments (1567.8ms)
Comments with pagy (188.0ms)
Processing by CommentsController#index as HTML
Comments (3659.9ms)
Comments with pagy (1039.4ms)
Processing by CommentsController#index as HTML
Comments (2545.5ms)
Comments with pagy (199.1ms)
Processing by CommentsController#index as HTML
Comments (643.9ms)
Comments with pagy (140.0ms)
Processing by CommentsController#index as HTML
Comments (2331.8ms)
Comments with pagy (103.9ms)
```

app/controllers/admin/comments/mark_as_spam_controller.rb
innatewonderer: Sorry didn't push the code earlier :) I am already convinced on member routes and `def spam/ham`

PR #288, opened 2018-11-08
`Pagy` with comments

innatewonderer:
Styling issues:
1. For admin user in `/comments` when you click page number on one table, since it's using `?page=2` in the URL, it reloads both tables :/ Need to test production behavior
recent_comments

2. `current` CSS doesn't render correctly. Instead, it pushes current digit up
```
	em, a:hover, span.current {
		text-decoration: none;
		color: $samo-orange;
	}
```
recent_comments

3. Says `prev` instead of previous. Must be a setting somewhere on `pagy`. Haven't found it yet.


innatewonderer:
Awesome! 

@ofsound would you mind taking a look?

app/controllers/comments_controller.rb
```diff
@@ -65,11 +65,11 @@ def find_comment
   end
 
   def set_comments_made
-    @comments_made = Comment.where(commenter_id: @user.id).public_or_private(display_private_comments?).page(params[:made_page])
+    @pagy_comments_made, @comments_made = pagy(Comment.where(commenter_id: @user.id).public_or_private(display_private_comments?))
```
sudara: @innatewonderer Will paginate in this case was not using `params[:page]` but `params[:made_page]` (as in comments made) — the problem is that we have 2 things paginating on this page, so we don't want 1 param to control them both. Can you look into how to use different param names for this instance?

ofsound:
> Awesome!
> 
> @ofsound would you mind taking a look?

@innatewonderer The existing CSS padding rules were being applied twice to the active number links because of the differences in how the new HTML was structured. I added a rule to remove it within a .pagy-nav class.

app/controllers/comments_controller.rb
innatewonderer: Hmmm, I thought I was doing that by naming one thing `@pagy`, and another one `@pagy_comments_made`... but you are saying that the actual query doesn't return only `comments made` comments? interesting 🤔 Let me look into it.

innatewonderer: Got it. Fixed.

PR #308, opened 2018-11-21
Add User Profiles

sudara:


sudara:
Verifying that users with assets_count 0 actually have no assets:

```irb(main):007:0> User.where(assets_count: 0).count
   (10.8ms)  SELECT COUNT(*) FROM `users` WHERE `users`.`assets_count` = 0
=> 18047
irb(main):008:0> User.left_outer_joins(:assets).where(assets: {id: nil}).count
   (48.3ms)  SELECT COUNT(*) FROM `users` LEFT OUTER JOIN `assets` ON `assets`.`user_id` = `users`.`id` WHERE `assets`.`id` IS NULL
=> 18049```

Issue #301, opened 2018-11-15
Only show "Other tracks that need your love" on mass_edit when appropriate 


sudara:
The point of "Other tracks you love" is to encourage others to edit track descriptions where they might have forgotten them.

However, we probably should have two changes here so that the partial is only shown when:

1. The user has more than 1 track (so if this is the first upload, don't show it)
2. And those tracks actually are missing descriptions

![white theme remove other tracks that need your love 2018-11-15 18-44-52](https://user-images.githubusercontent.com/472/48571183-8d083100-e906-11e8-938b-d65e192be19a.jpg)

innatewonderer:
I've noticed that we redirect a user to `/tracks/mass_edit?assets%5B%5D=107305` on Asset # Upload and Publish (or Don't publish). Is it intentional to cover cases when multiple records are uploaded at once? (is it even a feature?)

Should we redirect to `assets#edit` instead?

PR #312, opened 2018-11-23
I301 love thy tracks

innatewonderer:
https://github.com/sudara/alonetone/issues/301

- [x] `assets#edit` The user has more than 1 track (so if this is the first upload, don't show it)
- [x] `assets#mass_edit` (for one track) The user has more than 1 track (so if this is the first upload, don't show it)
- [x] add descriptionless block and css to `admin#edit`
- [x] `assets#mass_edit` for (true mass edit) multiple tracks view
- [x] Exclude current asset from the list (if present)

@ofsound I copy/pasted CSS from mass_edit block. Please give your blessing.

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -118,7 +118,12 @@ class AssetsController < ApplicationController
 
   def mass_edit
     redirect_to_default && (return false) unless logged_in? && (current_user.id == @user.id) || admin?
-    @descriptionless = @user.assets.descriptionless
+    # currently we redirect asset # publish to mass_edit with params["assets"]
+    if params["assets"].first && @user.assets.not_current(params["assets"].first).descriptionless.count > 0
+      @descriptionless = @user.assets.not_current(params["assets"].first).descriptionless
+    elsif @user.assets.descriptionless.count > 2
+      @descriptionless = @user.assets.descriptionless
+    end
     @assets = [@user.assets.where(id: params[:assets])].flatten if params[:assets] # expects comma seperated list of ids
     @assets = @user.assets unless @assets.present?
     render 'mass_edit_white' if white_theme_enabled?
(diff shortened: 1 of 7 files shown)


app/controllers/assets_controller.rb
```diff
@@ -111,14 +111,19 @@ def new
...
-    @descriptionless = @user.assets.descriptionless
+    @descriptionless = @user.assets.not_current(@asset.id).descriptionless if @user.assets.not_current(@asset.id).descriptionless.count > 1
     @allow_reupload = true
     render 'edit_white' if white_theme_enabled?
   end
 
   def mass_edit
     redirect_to_default && (return false) unless logged_in? && (current_user.id == @user.id) || admin?
-    @descriptionless = @user.assets.descriptionless
+    # currently we redirect asset # publish to mass_edit with params["assets"]
```
sudara: @innatewonderer any chance you can elaborate on this comment? I'm wondering if the app is doing something it shouldn't be doing. Is it related to the logic below?

innatewonderer: When you click `publish` for a new track upload, I'd expect it to redirect to `/assets/edit` but instead you get redirected to `/mass_edit`... Wasn't sure if it's a feature or a bug. Wasn't sure if there is a way to upload multiple tracks at once or a desire to have such ability :)

sudara: Ah good call. Yeah, one track should go to /edit, multiple should go to /mass_edit — right now you can upload a zip file — and actually if we change the file selector to a multi on the front end, you can select multiple there too. Drag and drop coming Jan 2019 care of Manfred.

sudara: I'm going to merge this for now, but ideally we would clean this up ASAP, as it makes logic in the controller action more difficult to understand. Maybe the "descriptionless" stuff can be extracted into a protected method or even tucked away into a model so that the controller can be more readable.

PR #324, opened 2018-12-04
Refactor AssetNotificationJob. 

### LINKED ISSUE #202: When user uploads more than 1 track, bundle the email notifications into 1 email
I'm about to upload 17 tracks in one go —  People will get notified 17 times that I uploaded, which is sorta crazy! 

Basically, people can upload a zip, which is the primary use case for this happening. We need to create a "bulk" version of the email notifier, remove the `after_create :notify_followers, if: :published?` model code and do an explicit notification in each case from the controller.

innatewonderer:
Instead of accepting single id, accept an array of ids and iterate over each title to support mass upload functionality

TESTED end to end locally

- [x] Test to make sure email generated correctly
- [x] Make sure single track upload works as expected

fixes #202

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -124,6 +124,9 @@ class AssetsController < ApplicationController
     end
     @assets = [@user.assets.where(id: params[:assets])].flatten if params[:assets] # expects comma seperated list of ids
     @assets = @user.assets unless @assets.present?
+
+    AssetNotificationJob.set(wait: 10.minutes).perform_now(asset_ids: @assets.map(&:id), user_id: @user.id)
+
     render 'mass_edit_white' if white_theme_enabled?
   end
 
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -125,7 +125,7 @@ class AssetsController < ApplicationController
     @assets = [@user.assets.where(id: params[:assets])].flatten if params[:assets] # expects comma seperated list of ids
     @assets = @user.assets unless @assets.present?
 
-    AssetNotificationJob.set(wait: 10.minutes).perform_now(asset_ids: @assets.map(&:id), user_id: @user.id)
+    AssetNotificationJob.set(wait: 10.minutes).perform_later(asset_ids: @assets.map(&:id), user_id: @user.id)
 
     render 'mass_edit_white' if white_theme_enabled?
   end
@@ -149,6 +149,7 @@ class AssetsController < ApplicationController
         flashes += "'#{CGI.escapeHTML asset.mp3_file_name}' failed to upload. Please double check that it's an Mp3.<br/>"
       end
     end
+
     if @playlist
       flash[:ok] = (flashes + "<br/>You had ID3 tags in place so we created an album for you").html_safe
       redirect_to edit_user_playlist_path(@user, @playlist)
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -125,7 +125,9 @@ class AssetsController < ApplicationController
     @assets = [@user.assets.where(id: params[:assets])].flatten if params[:assets] # expects comma seperated list of ids
     @assets = @user.assets unless @assets.present?
 
-    AssetNotificationJob.set(wait: 10.minutes).perform_later(asset_ids: @assets.map(&:id), user_id: @user.id)
+    @user.followers.select(&:wants_email?).each do |follower|
+      AssetNotificationJob.set(wait: 10.minutes).perform_later(asset_ids: @assets.map(&:id), user_id: follower.id)
+    end
 
     render 'mass_edit_white' if white_theme_enabled?
   end
--- a/app/jobs/asset_notification_job.rb
+++ b/app/jobs/asset_notification_job.rb
@@ -1,7 +1,7 @@
 class AssetNotificationJob < ActiveJob::Base
   queue_as :mailers
 
-  def perform(asset_ids: asset_ids, user_id: user_id)
+  def perform(asset_ids: , user_id: )
     assets = Asset.find_by(id: asset_ids)
     user = User.find_by(id: user_id)
     AssetNotification.upload_notification(assets, user).deliver_now if assets && user&.email
--- a/app/jobs/asset_notification_job.rb
+++ b/app/jobs/asset_notification_job.rb
@@ -1,8 +1,8 @@
 class AssetNotificationJob < ActiveJob::Base
   queue_as :mailers
 
-  def perform(asset_ids: , user_id: )
-    assets = Asset.find_by(id: asset_ids)
+  def perform(asset_ids:, user_id: )
+    assets = Asset.where(id: asset_ids)
     user = User.find_by(id: user_id)
     AssetNotification.upload_notification(assets, user).deliver_now if assets && user&.email
   end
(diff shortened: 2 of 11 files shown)


innatewonderer:
As I was doing local testing, I noticed that few things were happening:
1. When you are uploading a .zip file with multiple songs, if you click `Save` on one - you will be redirected immediately (without saving changes on other songs).
2. We cannot send notification during mass_edit step because at that point title hasn't been assigned yet, so a user would get notification looking like this:

```
From: support@yourapp.com
To: jzueva@gmail.com
Message-ID: <5c085a808c273_160693fd19cef4a0440619@omac-C02PC00LG3QR.local.mail>
Subject: [alonetone] 'innatewonderer' uploaded a new track!
Mime-Version: 1.0
Content-Type: text/plain;
 charset=UTF-8
Content-Transfer-Encoding: 7bit

Hi!

Innatewonderer just uploaded a track to alonetone! OMG!

Listen to
   on alonetone: https://yourapp.com/innatewonderer/tracks/107347
   on alonetone: https://yourapp.com/innatewonderer/tracks/107348
   on alonetone: https://yourapp.com/innatewonderer/tracks/107349
   on alonetone: https://yourapp.com/innatewonderer/tracks/107350

------------------------------

You are getting this email because you decided to follow Innatewonderer. You can change your mind by clicking clicking here: https://yourapp.com/unfollow/innatewonderer

You can also turn off all emails by unsubscribing here: https://yourapp.com/notifications/unsubscribe.
```

innatewonderer:
@sudara what do you think if we print the titles that are available (if available), hopefully since we are waiting 10 minutes to send the email - user will get enough time to update titles and descriptions ... but in case they do not, we provide a generic link to uploader's profile?

sudara:
@innatewonderer Sounds good, that's what we are doing now — because the permalinks can change with a title update, the best thing to do is actually to link to the id (not the permalink) to ensure forward compatibility as the artist adjusts

sudara:
@innatewonderer I'm wondering if it makes most sense to actually have 2 different jobs, one for single track and one for a batch. The wording seems like it would be significantly different...

Also, we could optionally (in another PR) detect if the whole batch is in a playlist and use the playlist title and just provide a link to it...

Also, to avoid having to test with zips, you could change the input field to allow multiple selections — that will be the case in january when manfred works on uploads.

app/mailers/asset_notification.rb
```diff
@@ -1,28 +1,34 @@
...
+    assets_array
+  end
+
   def user_link_for(asset)
     'https://' + Alonetone.url + '/' + asset.user.login
   end
 
   def play_link_for(asset)
-    user_link_for(asset) + '/tracks/' + asset.id.to_s
+    user_link_for(asset) + '/tracks/' + asset.permalink.to_s
```
sudara: This was using `id` for a reason, see my other comment :)

innatewonderer: Yep, reverting.

app/jobs/asset_notification_job.rb
```diff
@@ -4,6 +4,9 @@ class AssetNotificationJob < ActiveJob::Base
   def perform(asset_ids:, user_id: )
     assets = Asset.where(id: asset_ids)
     user = User.find_by(id: user_id)
-    AssetNotification.upload_notification(assets, user).deliver_now if assets && user&.email
+
+    return unless user&.email
```
innatewonderer: @sudara how about something like this?

innatewonderer:
Now sure how to fix the failing spec. Still thinking about it.

app/views/asset_notification/upload_notification.text.erb
```diff
@@ -1,12 +1,12 @@
 Hi!
 
-<%= @name.capitalize %> just uploaded a track to alonetone! <%= @exclamation %>!
+<%= @user.name.capitalize %> just uploaded a track to alonetone! <%= @exclamation %>!
 
-Listen to <%= @title %> on alonetone:
+You can listen to <%= @track.title %> track on alonetone:
```
sudara: Can we revert this back to the original wording?

app/models/asset.rb
```diff
@@ -144,7 +143,7 @@ def to_param
 
   def notify_followers
```
sudara: Is this method called any more?

app/mailers/asset_notification.rb
```diff
@@ -1,22 +1,38 @@
...
     @stop_following_link = stop_following_link
     @unsubscribe_link = unsubscribe_link
     @exclamation = %w[Sweet Yes Oooooh Alright Booya Yum Celebrate OMG].sample
-    mail subject: "[alonetone] '#{asset.user.name}' uploaded a new track!", to: email
+    mail subject: "[alonetone] '#{@user.name}' uploaded a new track!", to: @user.email
+  end
+
+  def upload_mass_notification(assets, user, _sent_at = Time.now)
+    @user = user
+    @tracks = generate_track_hash(assets)
```
sudara: Since `Track` is a different model, can we rename this to `@assets` to prevent confusion?

app/controllers/assets_controller.rb
```diff
@@ -116,6 +116,9 @@ def mass_edit
     end
     @assets = [@user.assets.where(id: params[:assets])].flatten if params[:assets] # expects comma seperated list of ids
     @assets = @user.assets unless @assets.present?
+
+    AssetNotificationJob.set(wait: 10.minutes).perform_later(asset_ids: @assets.map(&:id), user_id: @user.id)
```
sudara: Kind of confused by this — `@user` is the uploading user — but we want the notification to go to all the users in `user.followers` — the idea is that when I upload, the 25 people following me will each get an email. So this is where `Asset.notify_followers` should be called.

app/mailers/asset_notification.rb
```diff
@@ -1,22 +1,38 @@
...
   default from: Alonetone.email
 
-  def upload_notification(asset, email, _sent_at = Time.now)
-    @track = asset.name
-    @description = asset.description
-    @name = asset.user.name
-    @user = asset.user
-    @title = asset.title.present? ? asset.title : "new track"
+  def upload_notification(asset, follower_email, _sent_at = Time.now)
+    @asset_author = asset.user
```
sudara: I understand wanting to get away from the ambiguous `@user` in situations like this. How do you feel about `@artist` or `@musician`?

app/models/asset.rb
sudara: Oooh thanks!

spec/mailers/asset_notification_spec.rb
```diff
@@ -4,7 +4,7 @@
   fixtures :assets, :users
 
   describe "upload_notification" do
-    let(:mail) { AssetNotification.upload_notification(assets(:valid_mp3), users(:sudara).email) }
+    let(:mail) { AssetNotification.upload_notification(assets(:valid_mp3), users(:sudara)) }
```
sudara: This is causing the current spec failure on the branch.

spec/request/admin/comments_controller_spec.rb
```diff
@@ -24,6 +24,9 @@
   end
 
   describe "unspam individual comments" do
+    # asset id 1 is needed for comment's commentable to return true
+    # for is_deliverable?
+    let!(:asset) { assets(:valid_mp3) }
```
sudara: Nice catch!

sudara:
Finally deployed this today and it's causing this issue https://app.bugsnag.com/alonetone/alonetone/errors/5c6af7963b5854001968b3e9?event_id=5c6af7bd00359e2babdf0000&i=sk&m=fq

PR #336, opened 2018-12-06
I335 tracks on white theme

innatewonderer:
Fixes#335


innatewonderer:
Noticed that `columns.right` display a bit higher than the `columns.left`. 

fullscreen_12_5_18__10_32_pm

app/controllers/assets_controller.rb
```diff
@@ -23,10 +23,18 @@ def latest
     end
   end
 
+  def set_related_user_variables
+    all_user_tracks = Asset.for_user(@user.id)
```
sudara: We can get rid of this line and below instead of using `all_user_tracks` just do `@user.assets.hottest` etc...

db/schema.rb
```diff
@@ -74,6 +88,8 @@
     t.index ["commentable_id"], name: "index_comments_on_commentable_id"
     t.index ["commentable_type", "is_spam", "private"], name: "index_comments_on_commentable_type_and_is_spam_and_private"
     t.index ["commenter_id"], name: "index_comments_on_commenter_id"
+    t.index ["remote_ip"], name: "index_comments_on_remote_ip"
+    t.index ["user_agent"], name: "index_comments_on_user_agent"
```
sudara: Not quite sure where this stuff came from...

db/schema.rb
```diff
@@ -12,6 +12,20 @@
 
 ActiveRecord::Schema.define(version: 2018_12_04_011838) do
 
+  create_table "active_admin_comments", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
```
sudara: looks like this table needs to be dropped on your end?

app/views/assets/index.html.erb
```diff
@@ -12,44 +12,51 @@
...
-  </div>
-  </div>
+    <div id="share_user_latest" class="box">
+      <h2 class="box static">Embed Your Latest Music</h2>
+      <div class="static_content">
+        <h3>Embed a flash player by copying and pasting:</h3>
+
+        <%= text_area_tag "flash_player_#{@user.id.to_s}", "<embed src=\"#{FLASH_PLAYER}\" width=\"400\" height=\"140\" allowfullscreen=\"true\" allowscriptaccess=\"always\" flashvars=\"&file=#{user_url(@user.login, format: :xml)}&height=140&width=400&frontcolor=0x3C3C3C&backcolor=0xf3f3f3&lightcolor=0xFF944B&screencolor=0xFF944B&displaywidth=120&showdigits=false&showdownload=true\" />", rows: 3, cols: 50 %>
+        (customize how it looks by visiting <%= link_to 'the setup wizard', 'http://www.jeroenwijering.com/?page=wizard&example=2'%>)
+      </div>
```
sudara: We can actually get rid of this whole "Embed Your Latest Music" — it was for an old embeddable player from years ago!

sudara:
@innatewonderer Can you merge master in? I think we should be good to go after that!

PR #371, opened 2018-12-31
Get bio from user's profile rather than user (which obvi doesn't work)

innatewonderer:


sudara:
Thanks! Need all the help we can get on user admin stuff — still being attacked pretty hard by spammers...

PR #372, opened 2018-12-31
I359 add paranoia gem and functionality

### LINKED ISSUE #359: Paranoia
With the new admin interface coming and so much moderation happening on the platform, it's become critical that we protect against deleting users and their uploads.

Here's what I'm thinking to start:

* [x] Add paranoia to both `User` and `Assets` — these are the only two things we really care about restoring.
* [x] Remove `before_destroy :efficiently_destroy_relations` callback from `User` — we'll want to keep everything around until the user is really destroyed.
* [x] Remove `dependent: :destroy` from model relations that won't have paranoia like pic, profile, playlist, comments — Read through https://github.com/rubysherpas/paranoia#destroying-through-association-callbacks for detail on how dependent destroy works with paranoia
* [x] Move all the `efficiently_destroy_relations` logic to occur `after_really_destroy` and add logic to manually destroy the pic, profile, playlists, comments...
* [x] When a user is soft deleted, we will now need to soft delete each asset. And same when a user is restored, we'll need to hook into `after_restore` to undelete each asset they have.
* [x] We'll need to make sure other objects from deleted users and tracks don't show up for normal users. So, we'll need to modify queries for https://alonetone.com/playlists, the comments on the homepage, the favorites on the home page
  * [ ] ~When a regular user or guest browses to a deleted user's page, they get a 404. So I shouldn't be able to see https://alonetone.com/deleted_user or https://alonetone.com/deleted_user/comments or any of those sub pages. We'll likely need to hook into `find_user` for this.~ _Jenya: Comes out of the package._
* [ ] Moderators *will* need to be able to see deleted users — so we will need the following
  * [x] When a moderator browses to a deleted user's page, they see the page with an alert that the user was deleted on X day.
  * [x] New scope for users admin page, showing `with_deleted`
  * [x] Ability to "restore" deleted users from the admin interface
* [ ] We'll need to setup a cronjob that will "really destroy" all deleted users older than 30 days. So let's make `User.really_destroy_users_older_than_30_days` that performs that task, along with a spec and then then we can pop a cronjob into place with `rails r`

innatewonderer:
Closes #359


app/models/user.rb
```diff
@@ -16,24 +18,23 @@ class User < ActiveRecord::Base
   scope :geocoded,      -> { where(['users.lat != ""']).recent                        }
   scope :alpha,         -> { order('display_name ASC')                                }
 
-  # The before destroy has to be declared *before* has_manys
-  # This ensures User#efficiently_destroy_relations executes first
   before_create :make_first_user_admin
-  before_destroy :efficiently_destroy_relations
+  before_destroy :enqueue_real_destroy_job
```
innatewonderer: @sudara How about something like this to handle our `cleanup after 30 days` requirement? 

I can def do a sidekiq scheduled job but I feel like we don't delete so many users to have it run every day. Thoughts?

app/controllers/application_controller.rb
```diff
@@ -67,9 +67,16 @@ def not_found
...
   end
 
+  def user_identifying_params
+     params[:login] || params[:user_id] || params[:id]
+  end
+
   def find_user
-    login = params[:login] || params[:user_id] || params[:id]
-    @user = User.where(login: login).first || (current_user if %w[new favorites].include? action_name)
+    login = user_identifying_params
```
innatewonderer: @sudara I give up. I still get `Cognitive Complexity` failures from CodeClimate. Pls help :)

app/models/user.rb
sudara: This looks great. Users are definitely deleted daily.... we currently have quite the spam problem, but we are talking 10-20 at most!

app/controllers/application_controller.rb
sudara: Yeah, this is a `cognitively annoying` chunk of code. There are a few subtle things going on here that would probably be best broken out to be explicit, for example, when deleting users `params[:user_id]` is being manually passed to differentiate the user we want to delete and the `current_user`

This logic is old and needs to be updated. IMO the only responsible way to do that is to figure out when and where we use these different params and write the code to clearly communicate those things. It's not worth it to just please codeclimate.

For example, we are also saying "disregard the `login` param if the action is `new` or `favorites` — there's a reason for that but we'd have to lay that all out and maybe move that logic back into the appropriate controllers.

My bet is the end solution would be to have several methods that are explicitly named and called by the controllers that care. For example, `users_controller#destroy` could call `find_user_by_user_id_param` which would be located in its controller (if it actually makes sense to send that param still) and the favorites controller could have `set_user_to_current_user` or something...

innatewonderer:
alonetone_and__navigation_html_erb_ _alonetone

sudara:
Looks like we are still red and the db:seeds problem is only on this branch. Let me know if I can help!

Gemfile.lock
```diff
@@ -225,7 +226,9 @@ GEM
       activerecord
       kaminari-core (= 1.1.1)
     kaminari-core (1.1.1)
-    kramdown (1.17.0)
+    kramdown (2.1.0)
```
sudara: This should be bumped separately too, I guess.

Gemfile.lock
```diff
@@ -63,12 +63,13 @@ GEM
       io-like (~> 0.3.0)
     arel (9.0.0)
     ast (2.4.0)
-    authlogic (4.4.2)
-      activerecord (>= 4.2, < 5.3)
-      activesupport (>= 4.2, < 5.3)
+    authlogic (5.0.0)
```
sudara: Not sure if this was intentional, but this PR deals is open for the authlogic bump: https://github.com/sudara/alonetone/pull/413

Gemfile.lock
```diff
@@ -559,6 +568,7 @@ DEPENDENCIES
   sass-rails
   scrypt
   selenium-webdriver
+  shoulda-matchers (= 4.0.0.rc1)
```
sudara: I like to be conservative with adding dependencies, as each one contributes to complexity, cost of maintenance, performance, etc. I usually will ask things like: Are we using this gem right now? If so, which features? Can we get away without it or will it save us a lot of work in the future? Can we remove some other gem if we are going to add this one?

Gemfile.lock
sudara: That authlogic PR exibits the `db:seeds` problem, so un-bumping here should get tests passing again and we can deal with validations and the authlogic upgrade separately.

innatewonderer: Sounds good.

Gemfile
```diff
@@ -28,6 +28,7 @@ gem 'has_permalink'
 gem 'authlogic'
 gem 'scrypt' # for authlogic
 gem 'request_store' # for authlogic
+gem "paranoia", "~> 2.2"
```
sudara: Can we remove the [pessimistic constraint](https://guides.rubygems.org/patterns/#declaring_dependencies) here? 

The current version is 2.4.x but I usually have a policy of not locking gems in the Gemfile unless there is known problems with upgrading to a higher version (otherwise things like `bundle update` and dependabot will never update it, we'd have to remember to remove it later)

sudara:
It looks like paranoia is going to block upgrading to Rails 6, we should chat briefly about it: https://github.com/rubysherpas/paranoia/blob/core/paranoia.gemspec#L27

It looks like there's an issue for it:
https://github.com/rubysherpas/paranoia/pull/456

But actually paranoia doesn't look actively maintained in the moment:
https://github.com/rubysherpas/paranoia/issues/464

iggant:
actually you shouldn't use paranoia, as you read at the gem description: 
 

> Notice:
> 
> paranoia has some surprising behaviour (like overriding ActiveRecord's delete and destroy) and is not recommended for new projects. See discard's README for more details.
> 
> Paranoia will continue to accept bug fixes and support new versions of Rails but isn't accepting new features.

sudara:
@iggant Yeah, we saw this immediately after I posted the comment! Thanks! We will do the work manually/explicitly.

PR #379, opened 2019-01-04
Create database tables for ActiveStorage

Manfred:
Add the migration created by `rails active_storage:install` and update `schema.rb`.


sudara:
Yay!

PR #380, opened 2019-01-04
Update webpacker configuration and fix slow specs

Manfred:
Specs were running slow because Webpacker was re-compiling packs on every example for some controllers.

I re-ran `rails webpacker:install` and commited some of the suggested changes: upgrade to edge Webpacker, include Babel configuration, and webpacker.yml changes. That didn't really solve the problem, but seems like a useful change.

Then I saw a compilation error caused by missing source: `MorphSVGPlugin`, so I decided to just take that out. Follow and like animations still run, was this even still used anywhere?

Finally I noticed that some entries in .gitignore were duplicated so I sorted and de-duped them.


sudara:
Ahh, sorry for the confusion — `MorphSVGPlugin` is used, but since alonetone is open source and that plugin isn't — I've uploaded it to slack and you'll have to put it manually in place. The animations  should depend on it — it should be the the difference between clunky and interpolated animations.

Bit confused about the babel changes — we use a `.babelrc` which looks like it probably does most of what the new code is doing and is kinda tweaked for stimulus.... maybe we don't need the new code.

Happy to merge the rest of the changes!

Manfred:
I researched it a bit and the difference is that `.babelrc` is file relative. So it applies to the current directory and all directories underneath. You can use `babel.config.js` for project-wide configuration, so that only works from the project root. I like the existing version because it's shorter.

Manfred:
Alright, I have a better plan. Closing this PR.

Issue #193, opened 2018-08-26
Investigate production bundle size (>600k?!)


sudara:
Post Rails 7:

```
bundle exec bin/webpack --mode="production" 
bundle exec bin/webpack --profile --json > stats.json
yarn run webpack-bundle-analyzer stats.json public/packs/
rm stats.json
```

old:

```
bundle exec bin/webpack --mode="production"
cp public/packs/application-<hash>.js .
bundle exec bin/webpack --profile --json --mode="production" > stats.json
yarn run webpack-bundle-analyzer stats.json
whybundled stats.json 
```

https://webpack.github.io/analyse/#modules

sudara:
bn.js and elliptical were both being required because of old version vendered version of seed_random.js that required npm's `crypto` lib.

![webpack bundle analyzer 2018-12-26 21-44-18](https://user-images.githubusercontent.com/472/50456821-6c94b300-0957-11e9-92ff-2221cd739501.jpg)

sudara:
Before seed random fix:
```
  application-72caa8338adf3ef56235.js (805 KiB)
```

After:

```
  application-3f1fa05da728c3318f79.js (462 KiB)
```

sudara:
Bundle situation afterwards: 

<img width="1573" alt="image" src="https://user-images.githubusercontent.com/472/50456886-eb89eb80-0957-11e9-8381-df1a012e2c0a.png">

sudara:
Before in production:

![2018-12-26 21-56-31](https://user-images.githubusercontent.com/472/50457066-217b9f80-0959-11e9-8d44-0f120e56a205.jpg)

After in production:

![2018-12-26 22-02-14](https://user-images.githubusercontent.com/472/50457172-ed54ae80-0959-11e9-9ec5-8bc3bf0191ad.jpg)

sudara:
Changed gsap import style to get under 100kb zipped https://github.com/sudara/alonetone/commit/4005082f70940336d15190b8a0cf87b38a885456

![2018-12-26 22-39-49](https://user-images.githubusercontent.com/472/50457873-2d6a6000-095f-11e9-85ae-8136badcbd5e.jpg)

sudara:
rails6rc2
```
-rw-r--r--  1 sudara  staff   373K Dec  9 22:29 application-af7e003efad5bc86ad31.js
```

99.5kb as of Rails 6.1

sudara:
[...]

PR #381, opened 2019-01-05
Update webpacker configuration and fix slow specs

Manfred:
Specs were running slow because Webpacker was re-compiling packs on every example for some controllers.

I re-ran rails webpacker:install and commited some of the suggested webpacker.yml changes. That didn't really solve the problem, but seems like a useful change.

Then I saw a compilation error caused by missing source: MorphSVGPlugin, so I added a stub file to work around that. Added some documentation to the README about that.

Finally I noticed that some entries in .gitignore were duplicated so I sorted and de-duped them.


sudara:
Touching the file from seeds/setup seems to make more sense to me! 

Both Ben and I need `MorphSVGPlugin` in place to verify animations are sane (as does production which has it manually placed) — this branch essentially quietly writes over the file.

Thanks for rooting this issue out. I've been wondering why tests on travis have been slowing down significantly lately :D

.gitignore
```diff
@@ -1,44 +1,37 @@
...
+public/sitemap.xml
 public/source_files
 public/system
-public/sitemap.xml
-.ruby-version
-/public/packs
-/public/packs-test
-/node_modules
+public/system/*
+stats.json/public/packs
```
sudara: I think we just want `stats.json` here — this is used for https://github.com/sudara/alonetone/issues/193

sudara:
Awesome! Appreciating that cleaner .gitignore and thanks for finding and fixing that webpacker thing!

PR #382, opened 2019-01-05
Use transactional fixtures instead of Database Cleaner and minor spec suite cleanups.

Manfred:
* Use transactional fixtures instead of Database Cleaner to manage the test database.
* Set all fixtures for all tests (they don't get loaded until used and are then kept around using transactions).
* Move all helpers from `rails_helper.rb` to their own support files.
* Review all configuration in `spec_helper.rb` and `rails_helper.rb` and clean up where possible.
* Add some documentation _why_ certain things are there.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -94,7 +94,6 @@ group :test do
   gem 'rspec-rails', :require => false
   gem 'rspec-mocks', :require => false
   gem 'guard-rspec', :require => false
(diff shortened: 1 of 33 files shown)


sudara:
It ain't spring yet, but feeling pretty optimistic with all this cleaning going on :)

PR #383, opened 2019-01-06
Reduce and merge setup for Rails and dependencies in spec suite.

Manfred:
* Remove unnecessary FactoryBot setup from helper.
* Merge all callbacks so there is just one for each stage (e.g. `:suite` or `:example`)
* Setup and manage Active Job centrally and remove it from each of the spec files.
* Remove `render_views` calls from specs because was already configured for the entire suite.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -70,8 +70,6 @@ gem 'skylight'
 gem 'sidekiq'
 gem 'dalli'
 
-
-
 group :production do
   gem 'puma'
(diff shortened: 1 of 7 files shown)


Manfred:
Percy is failing because of non-stable order in showing "Currently Kicking Ass". I assume because the `created_at` dates are identical in the fixtures. Should I add order by `id` to fix that?

PR #388, opened 2019-01-11
Change database encoding to utf8mb4

Manfred:
MySQL supports 4 byte wide utf-8 when you use the utf8mb4 encoding for all string and text fields. It allows you to use emoji and a lot of Chinese characters which are outside of the regular Unicode planes.

~~On older versions (< 5.7), which I assume Alonetone uses, you have to redefine indices that go over 768 bytes in length by restricting the number of codepoints which are indexed in the string column.~~

Cool thing about `CONVERT TO CHARACTER SET` is that it actually converts all existing data in a table for you. Not really a problem in this case because utf8mb4 is a superset of utf8.

Finally you also have to upgrade the MySQL driver / connection encoding so you don't loose the data in transit. That means changing `database.yml` as show in the example.

This needs some serious offline testing against the production database to make sure we're not breaking any existing data.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -170,9 +170,9 @@ end
 #  artist           :string(255)
 #  bitrate          :integer
 #  comments_count   :integer          default(0)
-#  credits          :text(16777215)
-#  description      :text(16777215)
-#  description_html :text(16777215)
+#  credits          :text(4294967295)
+#  description      :text(4294967295)
+#  description_html :text(4294967295)
 #  favorites_count  :integer          default(0)
 #  genre            :string(255)
 #  hotness          :float(24)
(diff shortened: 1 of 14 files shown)


sudara:
We are using a percona build of mysql 5.7.21 

```
ii  percona-server-common-5.7             5.7.21-21-3.bionic                amd64        Percona Server database common files (e.g. /etc/mysql/my.cnf)
```

```
root@wurli:~# mysql -V
mysql  Ver 14.14 Distrib 5.7.21-21, for debian-linux-gnu (x86_64) using  7.0
```

Manfred:
Alright, hopefully that doesn't require dropping the indices. Running multiple versions of MySQL is kind of pain so I hope I can make that work.

sudara:
Tested this locally, seems happy and safe!

PR #391, opened 2019-01-15
Do not show comments by the same ip on main page

innatewonderer:
Add a check for commenter's ip to make sure it does not match comment's ip.

Fixes#369


app/models/comment.rb
```diff
@@ -10,6 +10,7 @@ class Comment < ActiveRecord::Base
   scope :last_5_public,      -> { on_track.only_public.limit(5).preload(commenter: :pic, commentable: { user: :pic }) }
   scope :made_between,       ->(start, finish) { where('comments.created_at BETWEEN ? AND ?', start, finish) }
   scope :to_other_members,   -> { where("commenter_id != user_id") }
+  scope :not_my_ip,          -> { joins(:commenter).where("users.current_login_ip != remote_ip") }
```
sudara: Would you be up to just add the extra .where condition to the `to_other_members` scope? It looks like they will only be used together / mean the same thing.

innatewonderer: I was thinking about it but was trying to stay under code climate's method complexity. I'll update `to_other_members` and what CC says.

PR #394, opened 2019-01-16
Modernize fixtures

Manfred:
* Define all relations between fixtures using their names instead of IDs
* Remove meaningless generated fixtures
* Use fixture helpers to reference them in specs
* Add fixtures for blobs and attachments


sudara:
Yay!

PR #396, opened 2019-01-17
Load Alonetone configuration earlier so it can drive framework configuration

Manfred:
* Add a value class that knows about valid configuration variables so we can iterate without leaving old configuration lying around.
* Use `Rails.configuration.alonetone` to read application specific configuration.
* Rename configuration variables and document them.
* Add some configuration (e.g. S3 region).
* Add configuration for Active Storage.

---

You can either redo the setup by copy-pasting `config/alonetone.example.yml` to `config/alonetone.yml` and move over all relevant configuration or you can boot a console and follow the upgrade instructions printed to the terminal. Note that instructions may be different for running console and specs.

Note that the `test:` section must be copied verbatim for the suite to run properly.


Manfred:
Just for clarity; in case someone finds this in the future. The new configuration code knows about deprecated configuration keys and automatically translates them. It never breaks but informs you of unused configuration keys.

screen shot 2019-02-17 at 09 48 34

PR #397, opened 2019-01-17
BugFix: `undefined method `follows_user_ids' for nil:NilClass`

innatewonderer:
Add a check if user is present to avoid error. Might want to have better ui if no tracks are available but will do for now. 
Not sure why this happened, except if a user hit url directly.
I confirmed that on white theme this link is hidden and on dark theme it's greyed out.
alonetone_radio__latest
alonetone_radio__popular

https://app.bugsnag.com/alonetone/alonetone/errors/5c3ec38a682ed9001a787ff1?event_id=5c3ec38a0032e7d49f080000&i=em&m=nw

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -79,9 +79,9 @@ class AssetsController < ApplicationController
   def radio
     params[:source] = (params[:source] || cookies[:radio] || 'latest')
     @channel = params[:source].humanize
-    if !logged_in? && ['those_you_follow', 'songs_you_have_not_heard', 'mangoz_shuffle'].include?(params[:source])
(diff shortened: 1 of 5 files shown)


innatewonderer:
Yes, I think so :/ Not sure if it's possible that it's hit by crawlers? or users who deleted their account but had a page open? Not sure. But it's another error that should no longer show up in Bugs :)

sudara:
@innatewonderer Gotcha!

It might be nicer to handle at the controller level by ensuring the request 404s vs. handling at the model level with a nil guard (which will return a 200 with no assets in the UI). We could 404 on https://alonetone.com/radio/those_you_follow and other user-only radio pages unless the user is logged in? And add a request spec...

Could be a good time to move the radio action out into its own controller and remove the `params[:source]` magic in favor of real actions, but I'll leave that up to you :D

PR #398, opened 2019-01-18
Add `pagy-overflow` to handle Pagy::Overflow exception

innatewonderer:
Pagy overflow will redirect to the last page rather than throwing an exception
More info:
https://github.com/ddnexus/pagy/blob/master/docs/extras/overflow.md


sudara:
Thanks for this! It's awesome that you are watching the bugsnags. Moar plz!

Issue #400, opened 2019-01-26
Add Akismet spam detection to user creation / updating


sudara:
We are still getting a lot of spammers. We are catching most of the content spam now (spammy tracks and comments) but we are not yet catching user / profile spam.

Akismet works great (unlike captcha and invisible_captcha), so let's use it to detect spammers by sending their profile info like bio, email address, website, name, ip to akismet (just like Asset and Comment do).

We need to answer a few questions like:

* What happens when someone signs up and is marked as spam immediately. Can they activate? 

* If a user joins, makes a thredded post, then updates their profile with spammy stuff and we mark the user as spam — what happens to their content? Is it still displayed?

* We probably want to 'shadow ban' users marked as spam. That is, the user is created, they can navigate, but their activity is not visible to the rest of the users. In a way, this makes it like paranoia's need of having deleted users remain in the db but not be displayed in the front end. However, we don't want to allow them to create content, such as upload or create forum posts.

* We need to add the ability to view spammed accounts in the admin area and quickly unmark them as spam.

Success conditions:

* [ ] No more spammy users on https://alonetone.com/users
* [ ] We didn't actually delete good users from production (important ;)
* [ ] A way to mark users as spam / not spam from admin
* [ ] No more spammy forum posts (these are coming from spammy users which shouldn't have the ability to post)

innatewonderer:
- [ ] Run akismet on users#create

innatewonderer:
Some thoughts, @sudara :
- Should we add Recaptcha or something during sign up to help us control spam volumes?
- I've been thinking about `User onboarding` flow. 

As a test, If a user is marked as spam, what if we DON'T activate the user? We display an error message with an admin@alonetone.com email address (or something) where an angry/confused user can email for questions. If it happens and an actual person emails us, then we adjust the logic. (If we find a spammed but REAL user in `/admin/users` - we also adjust the logic.
**Pros**:
- a fake user doesn't create fake content that we have to clean up. 
- hopefully, it deters that "bot" from attempting to create accounts with us in the future

**Cons**: 
- if a real user accidentally marked as spam and is not allowed to proceed, we might lose him/her forever 😢

innatewonderer:
As a test I want to run a spam check on all users locally, then spot check all that were detected as spam by Akismet. If we find that it's pretty accurate, then a Cons above might be less risky.

PR #426, opened 2019-02-10
Add is_spam column to user

innatewonderer:
Part of Issue#400

https://github.com/sudara/alonetone/issues/400

@sudara do you deploy migrations as separate PRs? I'm just used to it, so here you go :)


sudara:
Not usually, but happy to do it that way.

Our deploy process looks like:

1) Code refresh on disk
2) Migrate if needed
3) Server restart

PR #427, opened 2019-02-10
Add akismet spam check to user model

innatewonderer:


spec/request/users_controller_spec.rb
```diff
@@ -39,17 +39,34 @@
...
 
       expect(flash[:error]).to be_present
       expect(response).to render_template("users/new")
     end
+
+    it "should raise an error if user fails Akismet check" do
+      akismet_stub_response_spam
+      post "/users", params: { user: { login: 'bar', password: 'foo' } }
+
+      expect(flash[:error]).to be_present
```
sudara: `flash[:error]` looks to be present here for another reason like missing form fields (we aren't yet explicitly adding to the flashes when a user is marked as spam)

app/models/user.rb
```diff
@@ -114,6 +114,11 @@ class User < ActiveRecord::Base
   # will be removed along with /greenfield
   has_many :greenfield_posts, through: :assets
 
+  include Rakismet::Model
+  rakismet_attrs  author: proc { display_name },
+                  author_email: proc { email },
+                  user_ip: proc { current_login_ip }
```
sudara: Nice catch on the `user_ip`! One question though: not sure that we have access to `current_login_ip` before the record is saved, since it's a column that authlogic is managing. It's probably worth putting in a request spec showing that a request IP makes it through to the akismet call, since the IP is the #1 piece of info being used to judge spamminess and we had trouble with it before.

I think we'll need to add keys like `content` here which can be `profile.bio` and `permalink`

innatewonderer: Great piece of info. I'll confirm

innatewonderer:
> We need to add some way to list the spammy users and let us unspam them, otherwise we will be flying blind shipping the PR.

I already adding this functionality in the soft-delete branch [here](https://github.com/sudara/alonetone/pull/441/files#diff-fb73907e1b647d8b8613daa06a146f10R30)

LMK if you want it to be moved as part of this PR.

app/models/user.rb
innatewonderer: You are correct, how about passing it from the [controller](https://github.com/sudara/alonetone/pull/427/files#diff-4e05ad0d64e6100656b63ad1e78f32c5R123)?
```
  def user_params_with_ip
    user_params.merge(current_login_ip: request.remote_ip)
  end
```

innatewonderer:
alonetone

spec/request/users_controller_spec.rb
innatewonderer: +1

sudara:
> I already adding this functionality in the soft-delete branch here

Doesn't that PR add delete/restore functionality? We'd have to be able to unspam a user in production to ship and I don't think we want to couple the work to soft-delete.... Can we add a sub-level to the navigation so we can see all/spam/not spam (like with comments) and put a "unspam" button in when the user is spam and "spam" button in when the user is normal?

innatewonderer:
I think we thought of Akismet after we began paranoia/soft delete branch, that's why all of the admin functionality is already there. I'll close that PR #441 and move `/admin/users` work here.

sudara:
Just had another thought — if we don't send out activation emails on create, won't we have to send them out on "unspam" ?

innatewonderer:
@sudara What about work related to a user story "do not display any content by spam users"? Do we want to add it here as well? This will grow this PR by quite a bit, but logically it makes sense.

innatewonderer:
Will open a new PR off soft-deleted branch

PR #428, opened 2019-02-16
The validation adds accessors to the model for the confirmation attribute

innatewonderer:
Per Rails guide: https://api.rubyonrails.org/classes/ActiveModel/Validations/HelperMethods.html#method-i-validates_confirmation_of

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,4 +1,7 @@
 class User < ActiveRecord::Base
+  # validates :password, presence: true
+  # validates :password_confirmation, presence: true
+  validates_confirmation_of :password
   concerned_with :validation, :findability, :settings, :statistics
 
   store :settings


sudara:
Does this fix the spec on the paranoia branch?

authlogic is already adding accessors and validations behind the scenes, including one for password and password_confirmation, so I'm confused about what exactly is going on...

```
irb(main):001:0> me = User.new
=> #<User id: nil, login: nil, email: nil, salt: "", created_at: nil, updated_at: nil, admin: false, last_login_at: nil, crypted_password: "", assets_count: 0, display_name: nil, playlists_count: 0, listens_count: 0, itunes: nil, comments_count: 0, last_login_ip: nil, settings: {}, posts_count: 0, moderator: false, followers_count: 0, login_count: 0, current_login_at: nil, current_login_ip: nil, persistence_token: nil, perishable_token: nil, last_request_at: nil, bandwidth_used: 0, greenfield_enabled: false, use_old_theme: false>
irb(main):002:0> me.password = "hey"
=> "hey"
irb(main):003:0> me.password_confirmation = "hey"
=> "hey"
irb(main)
```

That being said, we are waiting on upgrading to 5.0 of authlogic and one of the breaking changes is that validations have to be explicitly in the model now (yay!)

https://github.com/binarylogic/authlogic/blob/master/doc/use_normal_rails_validation.md

Maybe there is some oddness with the older authlogic and we should upgrade and make validations explicit on this branch: https://github.com/sudara/alonetone/pull/413

innatewonderer:
It should be taking care of it, that's why the spec failing in Paranoia branch is failing [SOME times](https://travis-ci.com/sudara/alonetone/builds/101108935).
```
5.91s$ bundle exec rake db:seed
rake aborted!
ActiveModel::UnknownAttributeError: unknown attribute 'password_confirmation' for User.
/home/travis/build/sudara/alonetone/vendor/bundle/ruby/2.6.0/gems/activemodel-5.2.2/lib/active_model/attribute_assignment.rb:53:in `_assign_attribute'
/home/travis/build/sudara/alonetone/vendor/bundle/ruby/2.6.0/gems/activemodel-5.2.2/lib/active_model/attribute_assignment.rb:44:in `block in _assign_attributes'
/home/travis/build/sudara/alonetone/vendor/bundle/ruby/2.6.0/gems/activemodel-5.2.2/lib/active_model/attribute_assignment.rb:43:in `each'
```
I was testing locally with `rake db:setup` and it wouild fail locally every time without it. And it worked wit this. I'll look into authlogic and see if it takes care of it. If it helps us merge paranoia branch though, I say let's merge both and I can revert this change.

sudara:
I'd definitely want to know a "why" before merging a solution, in particular because it has to do with authentication and passwords. If we don't know the why, then it's possible the root solution isn't actually handled, or there will be side effects from the solution we choose, etc.

It looks like the last 3 builds had this problem on the paranoia branch: https://travis-ci.com/sudara/alonetone/branches

I looked at the commit where it first went red:
https://github.com/sudara/alonetone/commit/8966bf07cb32d3a13e50fb897571b43e8b70e3ac

It looks like authlogic was somehow bumped a major version in that merge commit. I didn't catch that when code reviewing the paranoia branch, but I'm going to go ahead and say that's why we are seeing an issue, so I'll close this and we can deal with it on that PR...

innatewonderer:
👍

PR #431, opened 2019-02-17
Replace buggy @shopify/draggable with sortablejs

sudara:


sudara:
image

sudara:
Sortable adds about 24k to the bundle pre-gzip (brotli)

sudara:
@ofsound This is good to go. I can merge it like this, or give you chance to look into some css stuff first, up to you! 

It looks like `sortable-ghost` is added as the class to the one being moved, maybe we can make it a bit more clear ala http://sortablejs.github.io/Sortable/ — not sure why we aren't seeing the tracks "swap places" like in the examples though...

sudara:
@ofsound Going to merge so things can move forward, but maybe you can put this on your list.

PR #432, opened 2019-02-18
Upgrade users class to prepare for authlogic upgrade

innatewonderer:
It is recommended in a [upgrade guide](https://github.com/binarylogic/authlogic/blob/master/doc/use_normal_rails_validation.md#instructions-for-440) to do this before actually merging the upgrade.


app/models/user.rb
```diff
@@ -1,11 +1,49 @@
...
+    },
+    length: { maximum: 100 },
+    uniqueness: {
+      case_sensitive: false,
+      if: :will_save_change_to_email?
+    }
+
+  validates :login,
+    format: {
+      with: /\A[a-z0-9]+\z/,
```
Manfred: I would probably use `/\A[[:alnum:]]+\z/` to support Cyrillic, Japanese, etc.

app/models/user.rb
```diff
@@ -1,11 +1,49 @@
 class User < ActiveRecord::Base
   concerned_with :validation, :findability, :settings, :statistics
 
+  validates :email,
+    format: {
+      with: /@/,
```
sudara: I was sort of curious why a better regexp wasn't recommended in the guide like `URI::MailTo::EMAIL_REGEXP` https://stackoverflow.com/a/49925333

sudara: Ahh I see, they give what was used in Authlogic 4. My vote would be to move to that or move to something standard like the EMAIL_REGEXP.

app/models/user.rb
sudara: I think `/\A[a-z0-9]+\z/` will drop uppercase characters and potentially not match at all when the string starts with an upper case character, which is a charge in behavior. 

This is what is being used currently: `/\A[a-zA-Z0-9_][a-zA-Z0-9\.+\-_@ ]+\z/`

I'm fine with `:alnum:` although admittedly wary about casually deciding to support cyrillic / japanese / etc...

Manfred: Ah, then you probably want some variant with `\w` because that's a shortcut for `a-zA-Z0-9_`.

innatewonderer: Would this be sufficient to replace current functionality without `casually adding support` for other languages:
```
/\A[\w\.+-@]+\z/
```
@sudara ^

sudara: @innatewonderer I think unless we escape the `-`  that will accept characters in the range of `+` to `@` which might not be what we want. See https://regex101.com

The old paperclip one allowed dash, plus, minus, @ — as long as they weren't the first character —and then the underscore. But I'm thinking we should go with the more restrictive `/\A\w\+\z/` — the added benefit is that the error message you copied and pasted is actually correct: "only letters and numbers"

innatewonderer: I added `[/\A\w\+\z/]`. We should ensure the error message says what it needs to say if anyone is trying to signup with incorrect characters. I can check/confirm it in an hour, but I think we can put it in another PR (so this one can be merged)

sudara:
🎉

PR #433, opened 2019-02-19
Compact Thredded initializer and some additional ground work surrounding user avatars

Manfred:
Move avatar URL logic the User and Pic model so we can use it directly in the Thredded configuration.

* Adds URL methods and predicate methods to test for existence to User and Pic.
* Adds a way to serve mocked uploaded files during request specs.
* Adds view helpers to generate a user avatar URL (will be used in upcoming PR).
* Adds `ImageVariant` to slowly move to a single truth for all image variant information (will be used more in upcoming PR).
* Compact Thredded initializer so it's a little more readable.


Manfred:
I tested this branch locally. It's probably a good idea to check the `pics` table in production to see if all of them have an `image/*` content-type before merging this change.

app/models/image_variant.rb
```diff
@@ -0,0 +1,38 @@
...
+    hq: [3000, 3000]
+  }.freeze
+
+  # Fit a variant inside these dimensions (width x height) to generate the variant.
+  def self.resize_to_fit(variant_name)
+    VARIANTS[variant_name.to_sym]
+  end
+
+  # Raises ArgumentError with explanation when the variant name does not exist.
+  def self.verify(variant_name)
```
sudara: What's the reasoning to have the error handling code here? It seems "fluffy" but maybe there's something I'm missing. For example, if we remove 25x25 variants, we can insure no place in the code is asking for variants of that size. If for some reason the code asks for an unlisted size, what's the benefit of the method blowing up in a "fancy" way vs. "normal"?

app/models/pic.rb
```diff
@@ -33,6 +33,16 @@ class Pic < ActiveRecord::Base
       "#{basename(attachment, style)}_#{style}"
     end
   end
+
+  # Returns true when this represents a usable picture.
+  def image_present?
+    pic_file_name.present? && pic_content_type.start_with?('image/')
```
sudara: Some more context around why this code will be needed might be nice. I'm guessing it's shimming for something coming?

config/initializers/thredded.rb
```diff
@@ -1,36 +1,22 @@
...
-# This is the column used to search for users' names if/when someone is @ mentioned.
+# Use User#login for the @mention syntax and to present the name of the user. Note that this column
+# must be unique.
 Thredded.user_name_column = :login
 
-# User display name method, by default thredded uses the user_name_column defined above
-# You may want to use :to_s or some more elaborate method
-#Thredded.user_display_name_method = :to_s
+# Configure the method name to fetch the current user.
+Thredded.current_user_method = :current_user
```
sudara: Thank you!

config/initializers/thredded.rb
```diff
@@ -1,36 +1,22 @@
...
-
-# User avatar URL. rb-gravatar gem is used by default:
-Thredded.avatar_url = ->(user) { user&.pic&.pic&.url(:album) || ActionController::Base.helpers.image_url('/default/o-pic_white.svg') }
-
-# ==> Permissions Configuration
-# By default, thredded uses a simple permission model, where all the users can post to all message boards,
-# and admins and moderators are determined by a flag on the users table.
+# Configure which URL to use for an avatar.
+Thredded.avatar_url = ->(user) {
+  user.avatar_url(variant: :album) || UsersHelper.no_avatar_path
```
sudara: `|| UsersHelper.no_avatar_path` is redundant here.

config/routes.rb
```diff
@@ -173,4 +173,11 @@
       get '/playlists/:id/:asset_id', :to => 'playlists#show', as: 'show_track_in_playlist'
     end
   end
+
+  if Rails.env.test?
+    # Paperclip serves its files from the public directory, something that does't work well
+    # with mocked uploaded files in feature specs. The files controller serves these files
+    # so we don't have to muck around in the Rack middleware stack to solve this.
+    get '/system/*path', to: 'test/files#show'
+  end
```
sudara: Interesting!

spec/models/user_spec.rb
```diff
@@ -56,6 +56,46 @@
...
+
+      it "knows the user does not have an avatar" do
+        expect(user.avatar_image_present?).to eq(false)
+      end
+
+      it "does not return a URL to a variant" do
+        expect(user.avatar_url(variant: :large)).to be_nil
+      end
+    end
+  end
```
sudara: Appreciating the coverage here.

app/models/image_variant.rb
Manfred: I guess I added this instinctively. It's mostly to help front-end focussed developers figure out what's going on when they have a typo in the variants. Otherwise they might get an exception from deep in the stack somewhere about something being `nil`. I can keep the code lean and remove this if you want.

app/models/pic.rb
Manfred: There was some defensive programming in the code `pic&.pic&.url` I believe. I was trying to replicate why this might be the case. I don't know why Paperclip would return `nil` in the second case. I could just change this to `pic.nil?`?

config/initializers/thredded.rb
Manfred: No, `user.avatar_url` will return `nil` when there is no avatar. See the spec below.

app/models/image_variant.rb
Manfred: One of the issue with Paperclip is that it will happily generate a URL to a non-existent 'style'.

```ruby
Pic.first.url(variant: :mew) #=>
  "https://d15x8imun0vkea.cloudfront.net/pics/1/construction_mew.jpg"
```

sudara: Ah gotcha! Usually I'd vote to keep things lean, but actually we've had a history of broken images and so on, so I'm happy for things to go a couple clicks more explicit. Thanks!

app/models/pic.rb
sudara: Gotcha, was just wondering why the content type was being checked. We'll be ok on production, it seems:

```
irb(main):005:0> Pic.where('pic_content_type not like "image/%"').count
=> 0
```

config/initializers/thredded.rb
sudara: Ah gotcha, I misread as `user_avatar_url` and was looking at the helper which has the same logic.

PR #436, opened 2019-02-21
Remove simple from playlist

ofsound:
I added the labels back on this branch, fixed some formatting after simple form was removed (to fix the save playlist button).


innatewonderer:
No more `simple_`... life just got a bit more complicated :D

PR #440, opened 2019-02-26
Add an array fix spec

innatewonderer:
Is this not failing for anyone else?
I'm wondering if it's a Ruby version? I'm running 2.5.1...

### DIFF
--- a/app/controllers/test/files_controller.rb
+++ b/app/controllers/test/files_controller.rb
@@ -25,7 +25,7 @@ module Test
     end
 
     def path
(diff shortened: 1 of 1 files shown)


Manfred:
Yes, open ended Range literals (i.e. `number..`) are new in Ruby 2.6. I believe that's what Alonetone is running on in production right? I don't mind changing this btw.

sudara:
Yeah, production is on 2.6.x but feel free to press merge Jeyna if you need ruby 2.5.x. support.

innatewonderer:
I need it right now for maybe few more hours :D till I start running 2.6 locally :D

PR #441, opened 2019-02-26
Soft delete. Closes #359

### LINKED ISSUE #359: Paranoia

innatewonderer:
### Completed as part of this feature PR:
- [x] Introduce SoftDeletion module that will replace paranoia
- [x] add SoftDelete to `assets`
- [x] add SoftDelete to `tracks`
- [x] add SoftDelete to `comments`
- [x] add SoftDelete to `feature`
- [x] add SoftDelete to `listens`
- [x] add SoftDelete to `playlists`
- [x] add SoftDelete to `posts`
- [x] add SoftDelete to `topics`
- [x] add SoftDelete to `tracks`
- [x] add SoftDelete to `user`
- [x] add controller action to `admin/users#delete`
- [x] add supporting specs for `admin/users#delete`
- [x] add controller action to `users#delete`
- [x] add supporting specs for `users#delete`
- [x] add model method to efficient soft delete all of the relations if a user is soft-deleted
- [x] add supporting spec for efficient soft deletion of all user's relations
- [x] add `admin/user#restore` action
- [x] add supporting specs for `admin/user#restore` action
- [x] add efficiently restore action for all of the user's relations if a user is restored
- [ ] add supporting specs for restore action
### User stories:
- as an `admin`, I would like to be able to see Deleted Users tab in `/admin/users` page
- as an `admin`, I would like to be able to soft-delete and restore users and its relations efficients from `/admin/users` tab
- as an `admin` I would like to be able to soft-delete a user from user's personal page (ex: `/users/sudara`)
- as a user, I would like to soft-delete myself and all of my dependencies from my personal page
- as an `admin` I want to be sure that in 30 days from the moment of soft-deletion, all users will be automatically deleted
- When visiting any of the pages on the site, no user should ever see any of the soft-deleted records (done through default_scope in SoftDelete) 
- as an `admin`, I would like to be able to see ALL of the users including soft-deleted in `admin/users`
- as an `admin`, I would like to be able to access soft-deleted user's page


sudara:
https://github.com/sudara/alonetone/pull/479 made me think that it might be useful as a part of this PR to make sure db:seeds creates a handful of soft deleted users in addition to having one in fixtures. That will give us a better chance to work with things and style things properly in development.

app/models/topic.rb
```diff
@@ -1,4 +1,6 @@
 class Topic < ApplicationRecord
+  include SoftDeletion
```
sudara: Topics are the old forum, only available on the old theme and I believe are no longer even able to be created. So we can leave them out of the equation.

app/models/user.rb
```diff
@@ -166,8 +168,30 @@ def avatar_url(variant:)
...
   end
 
+  def soft_delete_relations
+    efficiently_soft_delete_relations
+  end
+
   protected
 
+  def efficiently_soft_delete_relations
+    Listen.where(track_owner_id: id).map(&:soft_delete)
```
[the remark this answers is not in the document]
innatewonderer: good catch!

app/models/topic.rb
innatewonderer: My only reason to add it was because we efficiently destroy them on `before_destroy` in [`user.rb`](https://github.com/sudara/alonetone/blob/master/app/models/user.rb#L229)

What about `features`? Wasn't sure about them either.

sudara: Oooh yeah, features are also old data, only kept for data posterity, I need to suck it up and delete em!

app/views/admin/_navigation.html.erb
```diff
@@ -1,13 +1,11 @@
...
-  <li>Comments</li>
-    <ul>
-      <li><%= link_to 'All',  admin_comments_path %></li>
-      <li><%= link_to 'Spam',  admin_comments_path({filter_by: {is_spam: true}}) %> </li>
-      <li><%= link_to 'Not Spam', admin_comments_path({filter_by: {is_spam: false}}) %></li>
-    </ul>
-    <li><%= link_to "Secretz", "/secretz" %></li>
-    <li><%= link_to "Spam Stats", "https://akismet.com/web/1.0/user-stats.php?blog=alonetone.com&api_key=#{Rails.configuration.alonetone.rakismet_key}"%></li>
-    <li><%= link_to "Job Status", "/sidekiq" %></li>
+  <li><%= link_to 'Comments',  admin_comments_path %></li>
```
sudara: We probably want to keep the <ul> within <ul> structure, but @ofsound knows for sure, he's styling it right now.

innatewonderer: +1

app/concerns/soft_deletion.rb
```diff
@@ -0,0 +1,25 @@
...
+    def soft_deleted?
+      deleted_at != nil
+    end
+
+    # would like to be able to skip any validation
+    def restore
+      update(deleted_at: nil)
+    end
+  end
+end
```
Manfred: I would put model concerns in `app/model/concerns`.

app/javascript/controllers/admin/user_controller.js
```diff
@@ -0,0 +1,25 @@
...
+  showDeleteButtonTarget() {
+    this.restorebuttonTarget.style.display = 'none'
+    this.deletedbuttonTarget.style.display = 'block'
+    this.deletedvalueTarget.parentNode.style.color = "black"
+  }
+
+  showRestoreButtonTarget(){
+    this.restorebuttonTarget.style.display = 'block'
+    this.deletedbuttonTarget.style.display = 'none'
+    this.deletedvalueTarget.parentNode.style.color = "red"
```
Manfred: I believe technically it would be better to add and remove a class name from the elements so the styling and transition can be handled through CSS.

app/models/user.rb
```diff
@@ -221,13 +224,64 @@ def avatar_url(variant:)
     pic&.url(variant: variant)
   end
 
+  def deleted?
+    deleted_at != nil
+  end
+
+  def soft_delete_relations
+    efficiently_soft_delete_relations
```
Manfred: If you want all `deleted_at` times to line up you could do `efficiently_soft_delete_relations(deleted_at)`.

app/concerns/soft_deletion.rb
sudara: I created an issue to move the remaining files out of app/concerns https://github.com/sudara/alonetone/tree/master/app/concerns

app/models/user.rb
```diff
@@ -221,13 +224,64 @@ def avatar_url(variant:)
...
+    Comment.joins("INNER JOIN assets ON commentable_type = 'Asset' AND commentable_id = assets.id")
+           .joins('INNER JOIN users ON assets.user_id = users.id').where('users.id = ?', id).update_all(deleted_at: Time.now)
+
+    assets.update_all(deleted_at: Time.now)
+
+    %w[tracks playlists posts comments].each do |user_relation|
+      send(user_relation).update_all(deleted_at: Time.now)
+    end
+
+    true
```
sudara: We no longer need to return true, this is an artifact of Rails 4: `before_filters` that returned false could stop the callback chain. https://blog.bigbinary.com/2016/02/13/rails-5-does-not-halt-callback-chain-when-false-is-returned.html

app/models/user.rb
```diff
@@ -221,13 +224,64 @@ def avatar_url(variant:)
...
+    Track.with_deleted.joins(:asset).where(assets: { user_id: id }).update_all(deleted_at: nil)
+    Comment.with_deleted.joins("INNER JOIN assets ON commentable_type = 'Asset' AND commentable_id = assets.id")
+           .joins('INNER JOIN users ON assets.user_id = users.id').where('users.id = ?', id).update_all(deleted_at: nil)
+
+    Track.with_deleted.where(user_id: id).update_all(deleted_at: nil)
+    Playlist.with_deleted.where(user_id: id).update_all(deleted_at: nil)
+    Post.with_deleted.where(user_id: id).update_all(deleted_at: nil)
+    Comment.with_deleted.where(user_id: id).update_all(deleted_at: nil)
+
+    true
```
sudara: Can delete this one too!

db/migrate/20190226022413_add_deleted_at_to_models.rb
```diff
@@ -0,0 +1,11 @@
+class AddDeletedAtToModels < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :deleted_at, :datetime
+    add_column :assets, :deleted_at, :datetime
+    add_column :tracks, :deleted_at, :datetime
+    add_column :playlists, :deleted_at, :datetime
+    add_column :listens, :deleted_at, :datetime
+    add_column :comments, :deleted_at, :datetime
+    add_column :posts, :deleted_at, :datetime
```
sudara: Don't hate me (or do!) but along with `Topic` we can also ignore `Post` — both of these are for the old forum. Although it might be nice to delete stuff from the new forum eventually, we can figure that out some other time...

spec/fixtures/comments.yml
```diff
@@ -65,8 +65,8 @@ public_comment_on_asset_by_user:
   commentable: valid_mp3 (Asset)
   body: "Well friend, this is your best work."
   is_spam: false
-  user: sudara
-  commenter: arthur
+  user: arthur
+  commenter: sudara
```
sudara: These changes are why Percy is ❌ — If you go to the build https://percy.io/alonetone/alonetone/builds/1628431 and look at the originals (not the diff) you can see that the home page comment disappeared. This is probably due to the "don't show comment on track from same user/ip" logic we added. We need to comment from the fixtures showing in Percy, so maybe better to add fixtures for your particular use if you need the owner switched here!

innatewonderer: Ah I didn't realize it was using spec files to fill those pages. I'd expect something like seed.rb or something. Good to know.

app/javascript/controllers/admin/user_controller.js
innatewonderer: Agree, my intention here was to spell out any CSS needs first, then move it into .scss

db/migrate/20190226022413_add_deleted_at_to_models.rb
innatewonderer: +1

app/models/user.rb
innatewonderer: +1

spec/fixtures/listens.yml
```diff
@@ -0,0 +1,8 @@
+valid_listen:
+  listener: sudara
+  asset: valid_mp3
+  track_owner: arthur
+valid_listen:
```
Manfred: You can't have multiple records with the same key in the fixtures. I generally like to use the key do describe what it does or why it's there. So in this case maybe `arthur_listens_to_valid_mp3`.

innatewonderer: oops. copy/pasta

sudara:
"Let's do this like Brutus!" https://www.youtube.com/watch?v=5CIa6l60k8Y

PR #455, opened 2019-03-03
Rails6

sudara:


lib/paperclip_processors/mp3_paperclip_processor.rb
```diff
@@ -37,7 +37,7 @@ def copy_id3_tags_to_record(info)
         if value.respond_to?(:encode)
           value.encode(
             'UTF-8', invalid: :replace, undef: :replace, replace: "\ufffd"
-          ).mb_chars.normalize
+          ).unicode_normalize(:nfkc)
```
sudara: @Manfred Since you recently added this, just wanted to tag you in case you had preferences.

```
DEPRECATION WARNING: ActiveSupport::Multibyte::Chars#normalize is deprecated and will be removed from Rails 6.1. Use #unicode_normalize(:nfkc) instead. (called from block in copy_id3_tags_to_record at /home/travis/build/sudara/alonetone/lib/paperclip_processors/mp3_paperclip_processor.rb:40)
```

Manfred: Oh god, the scourge of the bad Ruby Unicode APIs continues. Yeah, it's fine to replace it like that. Make sure you include the `:nfkc` argument because the default is `:nfc` and that's a really ignorant default.

Gemfile
```diff
@@ -38,7 +38,7 @@ gem 'local_time'
 gem 'pagy'
 
 # deprecated
-gem 'record_tag_helper'
+gem 'record_tag_helper', github: 'mmoll/record_tag_helper', branch: 'update_gemspec'
```
Manfred: Maybe you want to copy the three methods this gem supplies to a helper file?

sudara: I like this idea a lot, thanks.

Gemfile
```diff
@@ -1,10 +1,10 @@
 source 'https://rubygems.org'
 
-gem 'rails', '~> 5.2.2'
+gem 'rails', '~> 6.0.0.beta2'
```
innatewonderer: Cutting edge :D

lib/paperclip_processors/mp3_paperclip_processor.rb
innatewonderer: omg this conversation making me so happy :awwww:

PR #465, opened 2019-03-11
Replace Paperclip.run with system

Manfred:
Might as well move `Greenfield::Waveform` into the application as `Waveform` and add a spec that it actually runs.


Manfred:
Why is CodeClimate triggering on those warnings when Rubocop doesn't locally?

sudara:
@manfred Rubocop runs in code climate as a plugin, but code climate has its own things it cares about. I don't consider it blocking. https://docs.codeclimate.com/docs/advanced-configuration#section-default-checks

app/models/asset.rb
```diff
@@ -154,7 +154,7 @@ def notify_followers
   end
 
   def create_waveform
-    Greenfield::WaveformExtractJob.perform_later(id)
+    WaveformExtractJob.perform_later(id)
```
sudara: This is pretty exciting to me, thanks!

PR #471, opened 2019-03-13
Complete removal of simple form

sudara:


sudara:
@ofsound Ok, this is ready for a little css love on 

* [ ] login form
* [ ] sign up form
* [ ] create new password form (after a password reset. You can reset password and look in the rails logs for the link to use to get to it)

One thing we have deal with is feedback when a field doesn't pass validation or is left empty. Do we want to display that inline with a field? Or as a list of "things wrong" at the top of the form? If you style some example I can plug it in, or I can just dump the errors out at the top....

app/views/playlists/edit.html.erb
```diff
@@ -16,10 +16,10 @@
 
         <%= render 'add_greenfield_downloads' if allow_greenfield_playlist_downloads?(@user) %>
 
-        <%= simple_form_for [@user, @playlist] do |f| %>
-          <%= f.input :title %>
-          <%= f.input :year, input_html: {maxlength: 4, size: 4} %>
-          <%= f.input  :description %>
+        <%= form_for [@user, @playlist] do |f| %>
```
innatewonderer: I thought that `form_tag` and `form_for` are soft deprecated since Rails 5... 1? 
Was there a reason for using `form_for` here?

sudara: form_for isn't yet deprecated, but you are right — if this was a template I expected to last more than another couple months, it should be changed.

ofsound:
@sudara I think we can close this branch? I updated the CSS for the ex-simple forms. For the failed form validations, I put an orange outline on the missing fields. I think that combined with the the flash message up top might be alright for now.

sudara:
@ofsound Perfecto — I looked into it and it would be pretty trivial to put errors next to the field, so we can see how it goes....

PR #473, opened 2019-03-14
Deal with warnings on Rails 6

Manfred:
Playlists are built instead of created when favoriting so they aren't saved in case of a validation error. This fixed a deprecation warning when trying to create a favorite through User.

Change call to `mb_chars.normalize` in asset spec and choose composed form because we're mostly interested in creating a normalized string regardless of how editors save the spec source.

### DIFF
--- a/app/models/track.rb
+++ b/app/models/track.rb
@@ -32,8 +32,9 @@ class Track < ActiveRecord::Base
   end
 
   def ensure_playlist_if_favorite
-    self.playlist_id = Playlist.favorites.where(user_id: user_id).first_or_create.id if is_favorite?
-    true
+    return unless is_favorite?
+
+    self.playlist = Playlist.favorites.where(user_id: user_id).first_or_initialize
   end
 end
 
--- a/spec/models/asset_spec.rb
+++ b/spec/models/asset_spec.rb
@@ -114,7 +114,7 @@ RSpec.describe Asset, type: :model do
     end
 
     it 'should handle umlauts and non english characters in the filename' do
-      filename = 'müppets.mp3'.mb_chars.normalize
+      filename = 'müppets.mp3'.unicode_normalize(:nfc)
       asset = file_fixture_asset(
         'muppets.mp3',
         filename: filename,
--- a/spec/models/track_spec.rb
+++ b/spec/models/track_spec.rb
@@ -14,12 +14,23 @@ RSpec.describe Track, type: :model do
   end
 
   context "as a fav" do
-    subject { users(:arthur).tracks.favorites.create(asset: assets(:valid_mp3)) }
+    let(:user) { users(:will_studd) }
+
     it "should create a favorite playlist if its the first fav" do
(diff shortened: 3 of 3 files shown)


spec/models/asset_spec.rb
```diff
@@ -114,7 +114,7 @@
     end
 
     it 'should handle umlauts and non english characters in the filename' do
-      filename = 'müppets.mp3'.mb_chars.normalize
+      filename = 'müppets.mp3'.unicode_normalize(:nfc)
```
sudara: I guess this changed in https://github.com/sudara/alonetone/commit/5242f8eebd3fdedaee8d209181159d4f2b0c59e6 but I'm kind of curious why we are manually normalizing at all here. Is it because of s3 needs?

app/models/track.rb
```diff
@@ -32,8 +32,9 @@ def set_user_id_from_playlist
   end
 
   def ensure_playlist_if_favorite
-    self.playlist_id = Playlist.favorites.where(user_id: user_id).first_or_create.id if is_favorite?
-    true
+    return unless is_favorite?
+
+    self.playlist = Playlist.favorites.where(user_id: user_id).first_or_initialize
```
sudara: Thank you! I looked at this, got distracted and merged Rails 6 :D

spec/models/asset_spec.rb
Manfred: An editor (or any other data processor like a database) may store Unicode in composed or decomposed form, so we can't rely on what we are getting from the source code to get a stable representation for a string.

```ruby
'é'.chars #=> ["é"]
'é'.unicode_normalize(:nfd).chars #=> ["e", "́"]
```

When you do a string comparison in a spec you will get a false negative if any processor force the string to another normalization form.

```ruby
'é'.unicode_normalize(:nfd) == 'é'.unicode_normalize(:nfc) #=> false
```

I can't remember if this was an issue here, but I assume that's why I added it.

Manfred:
@innatewonderer Oops, I was running around yesterday and didn't read your question until later. The deprecation warnings were:

```
DEPRECATION WARNING: Class level methods will no longer inherit scoping from
`create` in Rails 6.1. To continue using the scoped relation, pass it into the
block directly. To instead access the full set of models, as Rails 6.1 will,
use `Playlist.unscoped`. (called from ensure_playlist_if_favorite at
/Users/manfred/Code/alonetone/app/models/track.rb:35)
```

It doesn't really show the cause of the issue so I turned on `config.active_support.deprecation = :raise` in `environments/test.rb` to get the full trace to the issue. The second one you've probably seen:

```
DEPRECATION WARNING: ActiveSupport::Multibyte::Chars#normalize is deprecated
and will be removed from Rails 6.1. Use #unicode_normalize(:nfkc) instead.
(called from block (3 levels) in <top (required)> at
/Users/manfred/Code/alonetone/spec/models/asset_spec.rb:117)
```

PR #475, opened 2019-03-14
Add column to playlists to track low quality covers

Manfred:
Allows us to eventually remove the `pic_id` column currently used by Paperclip backed models when we switch to Active Storage.

### DIFF
--- a/app/helpers/playlists_helper.rb
+++ b/app/helpers/playlists_helper.rb
@@ -17,22 +17,12 @@ module PlaylistsHelper
     content_tag(:div, '', class: 'no_pic')
   end
 
-  # Returns true when the Pic with this ID does not have a greenfield variant.
-  def no_greenfield_variant?(pic_id)
-    (69806..72848).cover?(pic_id)
-  end
-
-  # Returns true when the Pic with this ID does not have a greenfield nor an original variant.
-  def no_greenfield_and_original_variant?(pic_id)
-    pic_id < 69807
-  end
-
   # Returns a different variant when the Pic with the supplied ID does not have the variant.
-  def downgrade_variant(pic_id, variant:)
-    if no_greenfield_and_original_variant?(pic_id)
+  def downgrade_variant(playlist, variant:)
+    if playlist.ancient_cover_quality?
       downgrade_ancient_variant(variant: variant)
-    elsif no_greenfield_variant?(pic_id)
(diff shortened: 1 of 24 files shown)


app/helpers/playlists_helper.rb
```diff
@@ -17,22 +17,12 @@ def playlist_cover_div
...
 
-  # Returns true when the Pic with this ID does not have a greenfield variant.
-  def no_greenfield_variant?(pic_id)
-    (69806..72848).cover?(pic_id)
-  end
-
-  # Returns true when the Pic with this ID does not have a greenfield nor an original variant.
-  def no_greenfield_and_original_variant?(pic_id)
-    pic_id < 69807
-  end
```
sudara: @Manfred I'm seeing problems in production related to this (not sure if this is cause) — for example check out all of this guy's playlists: https://alonetone.com/thetworegs — we don't seem to be scaling up the smaller image anymore.

db/migrate/20190313134953_downgrade_cover_quality_for_older_playlists.rb
```diff
@@ -0,0 +1,6 @@
+class DowngradeCoverQualityForOlderPlaylists < ActiveRecord::Migration[5.2]
+  def change
+    Playlist.where(pic_id: 0..69807).update(cover_quality: :ancient)
+    Playlist.where(pic_id: 69806..72848).update(cover_quality: :legacy)
+  end
+end
```
sudara: Somehow this didn't take, so a bunch of covers went to broken images.

```irb(main):003:0> me.id
=> 6689
irb(main):004:0> me.cover_quality
=> "modern"```

```
irb(main):002:0> Playlist.where(pic_id: 0..69807).count                          
=> 0
```

app/helpers/playlists_helper.rb
Manfred: Yes, I noticed it too with my own playlists. I'm not sure what is wrong here unless I can browse around in the production database. Maybe we can screen share later today and you can poke around a bit?

db/migrate/20190313134953_downgrade_cover_quality_for_older_playlists.rb
Manfred: Ah, you already fixed it. Sorry about this.

PR #479, opened 2019-03-19
More seeds

Manfred:
Current seeds are not really sufficient because they don't contain everything you would need to properly work on design.

Adds more users, playlists, assets, and listens. Introduces comments, favorites, and user profiles.


db/seeds.rb
```diff
@@ -40,22 +74,30 @@ def put_user_credentials(username, password)
   email: 'musician@example.com',
   password: musician_password,
   password_confirmation: musician_password,
-  greenfield_enabled: true
+  greenfield_enabled: true,
+  current_login_ip: 'fc00:1:1::3'
```
sudara: Hehe, what's this ipv6 addy?

Manfred: It's the equivalent of `10.0.1.3`, see: https://en.wikipedia.org/wiki/Unique_local_address

db/seeds.rb
```diff
@@ -1,23 +1,53 @@
+# Tell Raskimet we're in testing mode is it doesn't attempt to train based on
+# any API requests from the seeds.
+Rails.application.config.rakismet.test = true
```
sudara: Good thinking.

sudara:
@Manfred Really awesome. I'll exercise patience on that front, then.

PR #483, opened 2019-03-21
Copies all Paperclip attachments to Active Storage attachments

Manfred:
Defines Active Storage attachments for all models with Paperclip attachments. Adds a script to create Active Storage attachments for all existing uploads and copy the S3 objects to the new locations.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -34,6 +34,8 @@ class Asset < ApplicationRecord
     source: :user,
     through: :tracks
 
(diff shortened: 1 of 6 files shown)


sudara:
This is too exciting.

sudara:
Am I right in thinking merging this won't alter production behavior, just makes it possible to run this script when we want to start the migration?

Manfred:
Correct, this doesn't change any production code.

PR #484, opened 2019-03-24
Akismet on soft delete

innatewonderer:
This PR contains Akismet logic on a user model as well as the `.spam?` check during user#create (only).

### DIFF
--- a/app/controllers/profiles_controller.rb
+++ b/app/controllers/profiles_controller.rb
@@ -4,7 +4,7 @@ class ProfilesController < ApplicationController
   def update
     @user.profile.update(profile_params)
     if @user.spam?
-      @user.is_spam = @user.spam?
+      @user.is_spam = true
       # delete all user's associations
       @user.soft_delete_with_relations
       flash[:error] = "Hrm, robots marked you as spam. If this was done in error, please email support@alonetone.com and magic fairies will fix it right up."
--- a/app/controllers/profiles_controller.rb
+++ b/app/controllers/profiles_controller.rb
@@ -3,7 +3,15 @@ class ProfilesController < ApplicationController
 
   def update
     @user.profile.update(profile_params)
-    redirect_back(fallback_location: edit_user_path(@user))
+    if @user.spam?
+      @user.is_spam = @user.spam?
+      # delete all user's associations
+      @user.soft_delete_with_relations
+      flash[:error] = "Hrm, robots marked you as spam. If this was done in error, please email support@alonetone.com and magic fairies will fix it right up."
+      redirect_to logout_path
+    else
+      redirect_back(fallback_location: edit_user_path(@user))
+    end
   end
 
   protected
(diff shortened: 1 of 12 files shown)


sudara:
This looks gerat! But we need a way to "unspam" and correct Akismet (via the admin interface) before we can let the app mark things as spam I think...

sudara:
To keep the admin interface making sense, I think we need:

1) A "Only Spam" navigation item under "Users" (I think we want to keep "Deleted" because a user can be deleted for many reasons, not just spam).

2) A "Spam" button that marks the user as spam *and* deletes and then changes to an "Unspam" button that notifies akismet and restores. In other words, we can just always assume spamming/unspamming changes the deleted state of the user in the app.

image

innatewonderer:
Flow

Spam user
-> user.is_spam = true
-> user.soft_delete
-> user.soft_delete_relations
-> show RESTORE btn

Restore user
-> user.restore
-> user.restore_relations
-> show UNSPAM btn if user.is_spam?
-> show DELETE btn

Delete user
-> user.soft_delete
-> user.soft_delete_relations
-> show RESTORE btn

app/controllers/profiles_controller.rb
```diff
@@ -3,7 +3,15 @@ class ProfilesController < ApplicationController
 
   def update
     @user.profile.update(profile_params)
-    redirect_back(fallback_location: edit_user_path(@user))
+    if @user.spam?
+      @user.is_spam = @user.spam?
```
sudara: The `.spam?` call makes an API request here, so we'd probably want to avoid 2 requests and do
```suggestion
      @user.is_spam = true
```

app/controllers/users_controller.rb
```diff
@@ -30,8 +30,16 @@ def new
   end
 
   def create
-    @user = User.new(user_params)
-    if is_a_bot? && @user.valid?
+    @user = User.new(user_params_with_ip)
+
+    if @user.spam?
+      @user.is_spam = @user.spam?
```
sudara: And here

```suggestion
      @user.is_spam = true
```

app/controllers/profiles_controller.rb
sudara: Down the line, we could (should?) maybe make this a background jerb to avoid a network request in our request cycle. But we'd have to think through the logic on the UI side (maybe we show the flash[:error] on every request by the logged in user where `is_spam` is true?)

app/controllers/users_controller.rb
```diff
@@ -33,7 +33,7 @@ def create
     @user = User.new(user_params_with_ip)
 
     if @user.spam?
-      @user.is_spam = @user.spam?
+      @user.is_spam = true
```
innatewonderer: ah, I see... it makes the call every time you call `is_spam?`.
For some reason I assumed it'd get cached... Good catch.

PR #486, opened 2019-03-27
Explain how to install front-end dependencies

Manfred:
I was forced to reinstall my Mac today and I found that `yarn` is never mentioned in the README. This PR adds instructions about installing `yarn` when setting up the application.


sudara:
Thanks!

PR #524, opened 2019-04-12
User Playlists

sudara:
The goal of this branch is to get white theme playlist management sane.

Sudara

* [x] Limits the playlists on the user page (like `/sudara`) to only 4 playlists
* [x] Add a button to "view more" that goes to `/sudara/playlists`
* [x] When new playlists are added, they should show up first in the list
* [x] `/sudara/playlists` should be sortable for owner/moderator and otherwise just a grid
* [x] Instructions should show for owner/moderator
* [x] Verify only owner/moderator can sort in backend
* [x] Add `private` class to private playlists to aid with css
* [x] Sort Playlists button on /user for users with less than 4 public playlsts
* [ ] Feature spec that confirms reordering works

Ben

* [x] "reorder playlists" button
* [x] Use same logic as on homepage for playlist responsiveness (single row of 4, single row of 3, growing grid of 2x2)
* [x] CSS differentiation of private playlists. Overlay padlock icon top right of cover? Place the whole playlist on a solid background color with a border, same as the private comment?
* [x] Take a look at `user_playlists_area` and `home_playlists_area` and see if this new view should share any of that logic.
* [x]  Move the playlists back on top for mobile /user
* [ ] Confirm we're mobile friendly
* [x] Add lock icon to private playlists when owner/moderator.

Open questions:

* Distinctions between View All showing up for Visitor vs Manage/Sort Playlists link showing up for authorized user ( at the top of /user homepage )
* Should we include private playlists in the ordering, or exclude them and list them separately. My vote would be to include them in the ordering, but does it subvert expectations to set a private playlist as 3rd on their home and then have it not show up at all? **Yes, let's include them**
* Do we want a seperate view for `/albums` or `/mixes` vs. one combine view for `/playlists`? **Yes, but for now let's just have them combined, since there are very few mixes on alonetone**


sudara:
For a guest/user looking at someone else's page:

* 1-4 playlists: no need for button
* > 4 playlists: "view all"

For a user looking at their own page / moderator, we want to allow playlist sorting.

* 1 playlist (including private): no need for a button
* 2-4 playlists (including private): "reorder playlists"
* > 4 playlists: "reorder playlists"

sudara:
@ofsound Thinking about merging this... any remaining issues on your end?

sudara:
@ofsound We're showing private playlists on the user home now, but the lock icons aren't showing up there yet.

ofsound:
> @ofsound We're showing private playlists on the user home now, but the lock icons aren't showing up there yet.

Hey @sudara 

I've generalized the padlocks and label changes, so they show on user home.

I've also re-ordered the mobile view to show the 1-4 playlists up top on a user home instead of further down the page.

app/assets/stylesheets/white_theme/user.scss
```diff
@@ -74,6 +71,73 @@ main {
...
+      }
+      .below_box {
+        margin-top: -36px;
+        margin-bottom: 26px;
+        @media #{$mobile} {
+          margin-top: 0;
+        }
+        &.sparse {
+          text-align: left;
+        }
```
sudara: @ofsound Just FYI I talked with smoofles and he recommended putting the button on the left side when there's only 2/3 playlists. So I added some logic to do that, as well as this classname, and increased the bottom margin a bit...

PR #529, opened 2019-04-22
Fix legacy sitewidth padding

ofsound:
This branch fixed some bad horizontal paddings on /users and /playlists, so those will now collapse similarly to the rest of the site on overall browser width. I also removed those old large black headings to match the Orange headings on the rest of the site. Also, made a new white theme menu for /users. So yeah, sorta escaped the original scope of this branch. But it should be all non conflicting and positive changes.

### DIFF
--- a/app/views/playlists/all.html.erb
+++ b/app/views/playlists/all.html.erb
@@ -1,19 +1,23 @@
+<% if white_theme_enabled? %>
+
+  <div class="playlists_container">
+
+    <h2>Browse all alonetone playlists</h2>
+
+    <ul class="playlists">
+      <%= render partial: 'shared/playlist_white', collection: @playlists, as: :playlist %>
+    </ul>
+  </div>
+
+<% else %>
+
 <div id="static_header">
   <div class="title">
     <h1 class="static">Browse all alonetone playlists</h1>
   </div>
 </div>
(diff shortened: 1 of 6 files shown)


app/views/users/_users.html.erb
```diff
@@ -1,20 +1,47 @@
 <div id="user_index">
-  <div id="static_header">
-    <div class="title">
-      <h1 class="static">Browse all alonetone artists</h1>
+
+
+  <% if white_theme_enabled? %>
```
sudara: fancy footwork!

app/views/playlists/all.html.erb
```diff
@@ -1,24 +1,28 @@
...
 
+    <ul class="playlists">
+      <%= render partial: 'shared/playlist_white', collection: @playlists, as: :playlist %>
+    </ul>
   </div>
 
 <% else %>
+
+<div id="static_header">
+  <div class="title">
```
sudara: Is there a reason this `div` needs to be here, or can we toast it? Seems like a lot of nesting for an h1...

ofsound: I think that's just legacy for the old dark theme at this point. static_header and static_content are two labels that will go away at the dawn of the theme unity!

PR #526, opened 2019-04-21
Bug: User's comments on their tracks should not display on homepage 

innatewonderer:
Fixes #460 

## "Ready For Review" Checklist

* [x] PR title accurately summarizes changes
* [X] PR description
* [X] Does this fix any open issues?
    * Issue #460 
    * The scope `to_other_members ` should filter out comments where `user.id == commenter.id`  OR if comment's `remote_ip` address matches owner's `current_login_ip `
    * Admin will still see all comments on homepage.
* [X] What changes are you making?
Fix existing bug, where comments were not displaying on homepage accurately. Only comments by users to other users should be displayed.
Cleaning up assets/comments and user assets to be able to easily read comments on Percy reports.
* [X] Why did you implement it this way?
Scope existed already, just needed to fix the reference.
* [X] Is there anything reviewers need to know or pay attention to?
The scope itself is a bit confusing. 
* [X] Does anything special need to happen for deployment?
No
* [ ] ~Optional: Inline github PR comments explaining context of tricky / complicated / unexpected lines so reviewers can follow along.~
* [X] If appropriate, new tests were added for isolated methods or new endpoints
* [X] All tests green
* [x] Percy changes are purposeful or [explained](https://percy.io/alonetone/alonetone/builds/2079146?utm_campaign=alonetone&utm_content=alonetone&utm_source=github_status_public)
The first two pages' changes are due to new fixtures being added.
Fullscreen_6_23_19__12_48_AM

* [x] I booted up the branch locally, exercised any new code I wrote.
* [ ] ~If css was touched, I verified changes are happy on mobile (via Percy is ok)~
* [ ] ~I opened an issue for any logical followups~

PR #538, opened 2019-04-26
Validate spam user records before saving them

### LINKED ISSUE #535: User login is nil in admin/users#index
## Error in alonetone

**ActionView::Template::Error** in **admin/users#index**
No route matches {:action=>"show", :controller=>"users", :id=>""}, missing required keys: [:id]

[View on Bugsnag](https://app.bugsnag.com/alonetone/alonetone/errors/5cc1917e488f4c001811bf00?event_id=5cc1918e0038d308f9ef0000&i=gh&m=ci)

## Stacktrace

    app/views/admin/users/index.html.erb:35 - block in _app_views_admin_users_index_html_erb___3074819566922583342_69842781634420
    app/views/admin/users/index.html.erb:27 - _app_views_admin_users_index_html_erb___3074819566922583342_69842781634420

[View full stacktrace](https://app.bugsnag.com/alonetone/alonetone/errors/5cc1917e488f4c001811bf00?event_id=5cc1918e0038d308f9ef0000&i=gh&m=ci)

*Created by sudara via Bugsnag*

innatewonderer:
* [x] PR title accurately summarizes the changes
* [x] PR description
         <details>
         Fixes #535
         </details>

         <details>

         - Validate the user record before checking it for spam. 

         - Explicitly attempt to save spam user record to db
         </details>
         <details>

         - By validating user first (ensure that password/username etc are present) we can ensure that the error message is communicated if the user is a confirmed human. 

         - By saving spam user we have the ability to unspam it in the future if Akismet check is incorrect.

         - By explicitly saving the record we avoid any surprised (like finding that `.update_attribute` will save a record
         </details>
         <details>
         No
         </details>
* [ ] ~Optional: Inline github PR comments explaining context of tricky / complicated / unexpected lines.~
* [x] If appropriate, new tests were added for isolated methods or new endpoints.
* [x] All tests green
* [x] ~Percy changes are purposeful and/or explained~
* [x] I booted up the branch locally and exercised any new code I wrote
* [ ] ~Add issue for any needed followups~


innatewonderer:
@sudara this is a collection of some small bug fixes and begining of refactoring of user's controller (at least check for `valid?` before for `spam?`).
Feel free to merge at will.

app/controllers/users_controller.rb
```diff
@@ -32,21 +32,23 @@ def new
...
-      redirect_to login_url(already_joined: true)
-    elsif @user.valid? && @user.save_without_session_maintenance
-      @user.reset_perishable_token!
-      UserNotification.signup(@user).deliver_now
-      flash[:ok] = "We just sent you an email to '#{CGI.escapeHTML @user.email}'.<br/><br/>Just click the link in the email, and the hard work is over! <br/> Note: check your junk/spam inbox if you don't see a new email right away.".html_safe
-      redirect_to login_url(already_joined: true)
+    if @user.valid?
+      if @user.spam?
+        @user.is_spam = true
+        @user.soft_delete
```
sudara: Do we run into that same odd bug with `update_attributes` here, where the record will be saved?

innatewonderer: Well... Yes but I don't think it really matters since the user here is `valid?`
We could add this check after `bot?` and make `.save` explicit so at least it's obvious.
ThoughtS?

innatewonderer:
Percy was complaining about dates. Approved.

spec/controllers/profiles_controller_spec.rb
```diff
@@ -62,7 +62,7 @@
       it "should soft delete users playlists" do
         expect do
           put :update, params: { user_id: 'sudara', profile: { bio: 'very spammy bio' } }
-        end.to change(Comment, :count).by(-10)
+        end.to change(Comment, :count).by(-9)
```
innatewonderer: This is now also part of https://github.com/sudara/alonetone/pull/526
Will go away once we merge that PR

app/controllers/users_controller.rb
sudara: My only concern with is `.save_without_session_maintenence` is how we prefer to save users on create — this ensures the user is not logged in on create (as they have to verify their email).

sudara: But I haven't looked at that Authlogic stuff in a while, so.... maybe it's ok...

innatewonderer: But if we log out the user if the user is spam, doesn't it also ensures the user is not logged in?
If a user is spam, I don't think we need to even send an email to them. 
Thoughts? I'm going to find my user flow through #create action and paste it here.

sudara: Stepping back a bit....We are in `User#create` — do we want to create a user record at all if we're spam?

sudara: Right now, the code seems to rely on the bug being present (soft deleting a record that isn't yet created). Seems like if we want to go that route we should explicitly `save_without_session_maintenence` first to give us control over what's happening — if that bug is fixed, our code will still work.

innatewonderer: Blank_Diagram__Lucidchart

_not at all overkill_

app/controllers/users_controller.rb
```diff
@@ -31,22 +31,23 @@ def new
 
   def create
     @user = User.new(user_params_with_ip)
+    if @user.valid?
+      if @user.spam? && @user.save_without_session_maintenance
```
innatewonderer: @sudara How about this? 
We COULD move it inside the actual user.save_with_session_maintenance block on line 44 but I like being able to check for spam right away.
Confirmed that this action won't save the user if it's not spam :)
```
2.6.0 :001 > false && true
 => false
```

app/controllers/users_controller.rb
innatewonderer: I think we should create a spam user... what if user is not spam afterall? 
I pushed the `.save_without_session_maintenence` already... so that should answer this question? 

I was wondering if we can assign deleted_at time to user in the helper (rather than calling upadate_attributes), but it just doesn't seem... both necessary and "safe" :) How many times are you really going to be dealing with `User.new` object in your app...

sudara:
I've been doing a lot of work on spam lately, including fixing some of the akismet integration and communicating with akismet about best practices.  Check out http://alonetone.com/admin/users/not_spam and https://alonetone.com/admin/users/is_spam to see the current status, we are (almost) finally catching it all.

There's one question we should ask before proceeding with this branch — is the current behavior (which is unintentional/a bug) actually desired?

If we spam users on create and DON'T create a record, then we literally don't have a record of the creation attempt. That means that false positives (real users marked as spam) will not be possible to identify/quantify from the backend. For example, browsing https://alonetone.com/admin/users/is_spam — are we ok with that?

app/controllers/users_controller.rb
```diff
@@ -31,23 +31,22 @@ def new
...
-      redirect_to login_url(already_joined: true)
-    elsif @user.valid? && @user.save_without_session_maintenance
-      @user.profile.update_attribute :user_agent, request.user_agent
-      @user.reset_perishable_token!
-      UserNotification.signup(@user).deliver_now
-      flash[:ok] = "We just sent you an email to '#{CGI.escapeHTML @user.email}'.<br/><br/>Just click the link in the email, and the hard work is over! <br/> Note: check your junk/spam inbox if you don't see a new email right away.".html_safe
-      redirect_to login_url(already_joined: true)
+    if @user.valid?
+      if @user.spam? && @user.save_without_session_maintenance
+        @user.is_spam = true
```
sudara: This will need to be `update_attribute` or otherwise actually update the db explicitly.

sudara: In case we don't have one already, it would be useful to have a spec that demonstrates a user marked as spam get's saved properly as spam and then is considered soft deleted.

sudara:
This branch feels great. I appreciate the fact that you went the extra mile.

PR #540, opened 2019-04-26
[WIP] Issue 536 Refactoring concerns

### LINKED ISSUE #536: Remove app/models/user/*.rb and config/initializers/concerns.rb
It's no longer 2007. Time to remove these "faux" concerns.

These files were just "other buckets to put user logic in" so we can

1) Move some of it back to user.rb
2) Go through user.rb and clean up unused methods
3) If any clump of behavior is actually extractable into a testable module, we can do that.

innatewonderer:
Start concern cleanup.
Fixes #536


sudara:
We got concerns running in #576 but we could audit this PR against master to see what methods could be removed (due to concerns being modules now, it's too conflicty to roll with this PR)

PR #569, opened 2019-05-01
Cleanup Travis-CI configuration and install VIPS for image processing

Manfred:
* Use package syntax to install packages
* Just use yarn to install JavaScript dependencies
* Install VIPS from source because Active Storage requires a recent build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,33 @@ addons:
     packages:
     - libsndfile-dev
     - lame
+    # For VIPS
+    - libexpat1-dev
+    - gettext
+    - libglib2.0-dev
+    - liborc-0.4-dev
+    - libfftw3-dev
+    - liblcms2-dev
+    - libmagickwand-dev
+    - libopenexr-dev
+    - libcfitsio3-dev
+    - libgif-dev
+    - libgsf-1-dev
+    - libmatio-dev
+    - libopenslide-dev
+    - libpango1.0-dev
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,16 +5,38 @@ rvm:
...
 cache:
-- bundler
-- yarn
-- npm
+  bundler: true
+  yarn: true
+  directories:
+  - $HOME/vips
 before_install:
-- npm i -g node@latest
```
sudara: This used to be required, but seems like everything is happy now?

sudara:
Do we know what triggered master going red?

Are there any cons to libvips I should know about? I haven't done my homework there...

sudara:
I approved Percy, one diff was intermittent (date related) to be fixed and the other was intentional change from Ben on master.

.travis.yml
Manfred: Yeah, I think Ubuntu used to ship with an ancient version of Node. But probably good enough for now.

Manfred:
Thijs said the reason for the failure was trying to use NPM And Yarn simultaneously. Not really sure why that would cause libsass to stop building.

Cons for VIPS is that we need a recent version and there is no package on Ubuntu. Speed and less reliance on crufty ImageMagick is the primary reason for switching. The ImageProcessing gem used by Rails 6 can use either so we can always switch back.

PR #571, opened 2019-05-01
Run migration in 4 threads

Manfred:
Parallel defaults to processes, which seemed to break locally for me. I assume that would also break in production so let's force threads. I chose 4 because each thread will attempt to create a database connection and those are probably limited.


sudara:
Production has the pool set to 50, and only uses 6-8 I think, so good headroom there.

PR #572, opened 2019-05-02
Replace Paperclip with Active Storage

Manfred:
PR switches the entire application to use Active Storage and adds additional code to handle file uploads and downloads. It removes all code that attempts to use Paperclip attachments.

* Switches from ImageMagic to VIPS for image processing (to increase processing speed)
* Changes assets controller to accept asset attachment
* Changes playlists controller to accept cover image attachment
* Changes users controller to accept avatar image attachment
* Introduces classes to deal with URL locations for storage objects (`Location` and `CloudFrontLocation`)
* Changes controller methods to use new location classes to format a URL
* Changes `ImageVariant` to define al image variants
* Changes upload code to rely entirely on its metadata extraction.

Merge according to instructions on https://github.com/sudara/alonetone/issues/166.

PR #577, opened 2019-05-03
Remove Paperclip tables and columns and clean up preloads

Manfred:
Followup on https://github.com/sudara/alonetone/pull/572.

* Removes `pics` table.
* Remove `playlists.pic_id` column.
* Remove Pic model.
* Remove last few associations to the Pic model
* Write scope for all models involving attachments so they can be queried efficiently.


sudara:
Sorry for the percy noise, that's top of my list.

PR #580, opened 2019-05-05
Mass edit assets

### LINKED ISSUE #531: On mass_edit, no feedback on save button is disconcerting.
We want to use the exact same pattern (and UI) for saving on `mass_edit` that we do with comments.

This will probably mean creating a stimulus controller like `comment_controller.js.` We can coordinate with @ofsound on the PR to make sure styles are reusable.

### LINKED ISSUE #533: If track is caught in spam filter after edit, we need a UI warning to be loud and clear, contain instructions.

### LINKED ISSUE #553: Bulk Track Notification email should send on track creation, not on mass_edit
Just got this in the mail today.

alonetone  'montgomeru' uploaded new tracks! - sudara@modernthings net - Modern Things Mail 2019-04-27 18-54-36

Basically, it's an email notification that ben re-uploaded all his tracks, which he didn't do. What he did was visit mass_edit.

It looks like we are sending out the bulk notification on a GET visit to mass_edit, which is undesirable. We want to send the email out shortly after upload, but give enough time for people to fill out mass_edit...

Also, some of the wording here needs love — it shouldn't be "a track" —  it might be nice to say how many tracks were uploaded to visually distinguish it in the upload list. We should remove the single quotes around 'montgomeru' in the subject in both this email and the normal one too...

innatewonderer:
Fixes #553
Fixes  #533
Fixes #531

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -128,9 +128,10 @@ class AssetsController < ApplicationController
     end
     @assets = [@user.assets.where(id: params[:assets])].flatten if params[:assets] # expects comma seperated list of ids
     @assets = @user.assets unless @assets.present?
-
-    @user.followers.select(&:wants_email?).each do |follower|
-      AssetNotificationJob.set(wait: 10.minutes).perform_later(asset_ids: @assets.map(&:id), user_id: follower.id)
+    if params[:send_email]
+      @user.followers.select(&:wants_email?).each do |follower|
+        AssetNotificationJob.set(wait: 10.minutes).perform_later(asset_ids: @assets.map(&:id), user_id: follower.id)
+      end
     end
 
     render 'mass_edit_white' if white_theme_enabled?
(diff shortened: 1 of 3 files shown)


innatewonderer:
Specs for change in `assets#create -> spam` are coming

app/controllers/assets_controller.rb
```diff
@@ -176,15 +176,12 @@ def update
     @asset.update_attribute(:is_spam, @asset.spam?) # makes an api call
     @asset.publish! if params[:commit] == 'Publish'
 
-    if request.xhr?
-      result ? head(:ok) : head(:bad_request)
+    if result
```
innatewonderer: @sudara this will be fixing #531. 
I think since we are handling all of our flash messages through `flash[:]`, it might be good to stay consistent? Could def add stimulusjs if you'd prefer to start moving messages to it.

sudara: The forms on `mass_edit` are currently calling `update` via ajax and saving in place. We want to keep that behavior, as the point of the page is to be able to dance around your tracks and update descriptions. 

So we might want to address https://github.com/sudara/alonetone/issues/531 separately via adding a stimulus controller and a UX pattern similar to comments (spinner + confirmation that save happened). 

Alternatively, since this comment we added a stimulus controller called `flash_controller.js` for ajax flash messages, we could use that too.

innatewonderer: Got it

sudara:
Closing this in favor of #617.

PR #634, opened 2019-07-11
Alright v2

ofsound:
I just added a second @font-face for the  'Alright-v2-Normal-Regular' that tells the browser to use the Alright-v2-Normal-Bold.woff2 for bold weights.

This way I won't have to rewrite the CSS heading styles. Many of our 'bold' text comes from the default bold stylings on h1, h2. I would have to make a reset to stop the system from doing that. I really don't know which way is more semantically correct, but at least this way, nothing should change from how it was before.

* [ ] PR description
  * Does anything special need to happen for deployment?
* [ ] Optional: Inline github PR comments explaining context of tricky / complicated / unexpected lines so reviewers can follow along.
* [ ] If appropriate, new tests were added for isolated methods or new endpoints
* [ ] All tests green
* [ ] I booted up the branch locally, exercised any new code I wrote.
* [ ] If css was touched, I verified changes are happy on mobile (via Percy is ok)

### DIFF
--- a/app/assets/stylesheets/white_theme/variables.scss
+++ b/app/assets/stylesheets/white_theme/variables.scss
@@ -38,7 +38,7 @@ $one-column: 'only screen and (max-width: 748px)';
 
 @font-face {
   font-family: 'Alright-v2-Normal-Medium';
-  src: url('https://cdn.alonetone.com/fonts/Alright-v2-Normal/Alright-v2-Normal-Medium.woff2') format('woff2');
+  src: url('https://cdn.alonetone.com/fonts/Alright-v2-Normal-Medium.woff2') format('woff2');
   font-weight: normal;
 }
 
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/white_theme/variables.scss
```diff
@@ -30,10 +30,33 @@ $two-column-narrow: 'only screen and (min-width: 749px) and (max-width: 920px)';
 $one-column: 'only screen and (max-width: 748px)';
 
 // fonts
+@font-face {
+  font-family: 'Alright-v2-Normal-Regular';
```
sudara: Maybe we should move away from an individual weight in the name:

```suggestion
  font-family: 'Alright-v2-Normal';
```

sudara:
@ofsound Looking at Percy, it looks like the black font is having a bit of trouble. Maybe it's being double bolded? Maybe we should include it with the regular and bold as weight 700/800?

app/assets/stylesheets/white_theme/variables.scss
sudara: @ofsound thoughts?

PR #636, opened 2019-07-14
Add soft-delete and delete functionality on asset

innatewonderer:
### Related Bugsnag issues:
- [this](https://app.bugsnag.com/alonetone/alonetone/errors/5cdb9b90ba8d770019f8d20d?filters[event.since][0]=30d&filters[error.status][0]=open). Fixed by correcting [delete_path in the view](https://github.com/sudara/alonetone/pull/636/files#diff-b995a0cb36fb14b69efc706503fbdd8fR39)
### What this PR is About:
We want to be able to soft-delete assets in `/admin` namespace. In order to have deletion logic consistent through the app, move all deletion logic to one place (`/app/commands/asset_command`). 
This affected:
- admin/assets_controller
- assets_controller#destroy action
Currently, the button on click do the following: 
- `SpamButton` 
-- will mark a record `is_spam: true`
-- will notify Akismet that a record is spam
- `DeleteButton`
-- delete assets relations (comments, tracks, listens)
-- CountDown any playlist's tracks_count
-- soft-delete asset itself
- `RestoreButton`
-- restore asset
-- restore its relations (comments, tracks, listens)
-- CountUp any playlists' tracks_count` (through tracks)

(Existing functionality, nothing changed) A worker will really_destroy all soft_deleted assets per schedule (`def destroy_deleted_accounts_older_than_30_days`)

*Important Note*: I noticed that we actually do not have unspam button in the view. I could add it, however, follow up PR #672 changes the logic of how a record is `Spammed/SoftDeleted` and then `Restored` so not sure if it's worth adding the Unspam button at this point.

### Changes:
- add commands to handle 
-- soft_delete_relations
-- destroy_with_relations
-- restore_with_relations
- add the following buttons to `views/admin/assets/index` page:
-- delete
-- restore

#### All assets view

alonetone

Spam record will be soft-deleted in #673 

#### Spam assets view

alonetone

- refactor how asset's relationship are soft and hard deleted
- added stimulus js controller for admin/assets to handle Spam, Delete and Restore buttons.

- [x] Followup work: soft delete any spam assets #672 
- [x] all new changes have supporting specs
- [x] confirmed manually that all buttons actually Spam/Delete or restore record and its relationships


app/models/concerns/soft_deletion.rb
```diff
@@ -6,7 +6,7 @@ module SoftDeletion
 
     scope :only_deleted, -> { unscope(where: :deleted_at).where.not(deleted_at: nil) }
     scope :with_deleted, -> { unscope(where: :deleted_at) }
-
+    scope :destroyable,  -> { only_deleted.where('deleted_at < ?', 30.days.ago) }
```
sudara: Nice addition.

app/javascript/controllers/admin/asset_controller.js
```diff
@@ -0,0 +1,50 @@
...
+      this.showDeleteButton()
+    } else {
+      this.showSpamButton()
+      this.showDeleteButton()
+    }
+  }
+
+  connect() {
+    // console.log("Hello, Stimulus!", this.element)
+  }
```
sudara: Let's remove this method

app/javascript/controllers/admin/asset_controller.js
```diff
@@ -0,0 +1,50 @@
...
+
+  //////////////////////////////////
+  // SPAM button
+  //////////////////////////////////
```
sudara: We'll have to remove these comment blocks to please our current linter settings. If you don't want to use eslint locally I can see if code climate will run eslint. Re: these particular comments,  "SpamButton" is clear in the name of the methods, so I think we can toss em! Generally, I love comments, but prefer to reserve for the "why" or rationale of otherwise confusing things.

app/models/asset.rb
```diff
@@ -184,6 +184,12 @@ def role
     end
   end
 
+  def self.destroy_deleted_accounts_older_than_30_days
+    Asset.destroyable.find_each do |asset|
+      Asset::DestroyWithRelations.new(asset: asset).call
```
sudara: Kind of curious if we have other options with regards to syntax here. This is be a bit "chewy" for me.

app/commands/assets/soft_delete_relations.rb
```diff
@@ -0,0 +1,17 @@
...
+    def initialize(asset:)
+      @asset = asset
+    end
+
+    def call
+      asset.comments&.map(&:soft_delete)
+      # first update playlists
+      asset.playlists.update_all(['tracks_count = tracks_count - 1, playlists.updated_at = ?', Time.now]) unless asset.playlists.empty?
+      asset.tracks&.map(&:soft_delete)
+      asset.listens&.map(&:soft_delete)
```
sudara: `&.map(&:soft_delete)` will issue 1 sql call per relation, which is unideal, even if in a job vs. request cycle. We either need to devise a collection method in `soft_deletion.rb` or manually call `.update_all(deleted_at: time)`.

app/models/asset.rb
innatewonderer: Would you elaborate? :)

app/commands/assets/soft_delete_relations.rb
innatewonderer: Hmmm I guess my intention here was to be able to trigger any associated actions for `listens` or `tracks`. But since these things don't seem to have associations we might want to delete, then I think it's safe to just `update_all()`

innatewonderer:
> Nice, this stuff is looking solid! 💣
> 
> Deja vu from the user soft deletion — if we deploy this then "delete" will mean something different on the track edit page vs. admin page which is maybe dangerously confusing for mods. So we should update `AssetsController#destroy` to use soft delete as well.

I didn't want to pile all in one PR :) But absolutely can.

app/controllers/admin/assets_controller.rb
```diff
@@ -13,28 +13,30 @@ def index
...
       assets = scope.not_spam
 
       assets.map(&:spam!)
       assets.update_all(is_spam: true)
-      redirect_back(fallback_location: root_path)
+    end
+
+    def delete
+      Assets::SoftDeleteRelations.new(asset: @asset).execute
+      @asset.soft_delete
```
sudara: This is non-blocking, as it's only going to apply to these 2 locations and console usage, but it's worth noting we'll never want to call these 2 things separately, as that would leave the Asset in a strange state — it would be better to have 1 method that we call here, similar to `User#soft_delete_with_relations` or `User.spam_and_mark_for_deletion` (or potentially using `SoftDeleteAssetWithRelations` as the command)

app/commands/assets/soft_delete_relations.rb
```diff
@@ -0,0 +1,18 @@
+module Assets
```
sudara: It feels preemptive to namespace commands to me (it only adds noise at this point) since it's our first command. It might be cleaner to do something like https://github.com/nslocum/design-patterns-in-ruby/blob/master/commands/create_file.rb and call this `SoftDeleteAssetWithRelations`

app/commands/assets/soft_delete_relations.rb
```diff
@@ -0,0 +1,18 @@
...
+  class SoftDeleteRelations
+    attr_reader :asset
+
+    def initialize(asset:)
```
sudara: Keyword arguments are great when passing in options or needing to manage multiple arguments without having to worry about order. 

However, I'd prefer just to have a simple argument here, since we are just executing the command on the passed in object.

spec/fixtures/assets.yml
```diff
@@ -133,3 +133,14 @@ jamie_kiesl_the_duck:
   mp3_file_size: <%= 2.3.megabytes %>
   length: 140
   created_at: 2019-01-01 13:00:03
+
+asset_with_relations_for_soft_delete:
+  mp3_file_name: DeleteMeSoftly.mp3
```
sudara: https://www.youtube.com/watch?v=oKOtzIo-uYw 😆

app/controllers/admin/assets_controller.rb
```diff
@@ -13,28 +13,30 @@ def index
     def unspam
       @asset.ham!
       @asset.update_attribute :is_spam, false
-      flash.notice = "Track was made public"
-      redirect_back(fallback_location: root_path)
     end
 
     def spam
```
sudara: We probably need to `soft_delete` on spam and `restore` on unspam like we do with users and have appropriate behavior in the frontend to match — this could be a followup with a migration to soft delete the existing tracks marked as spam if you don't want to do it here.

sudara:
We might need to discuss this complicating things with asset soft deletion — https://github.com/sudara/alonetone/issues/653

app/controllers/admin/assets_controller.rb
innatewonderer: Makes sense. i'll update.

app/commands/assets/soft_delete_relations.rb
innatewonderer: My reasoning is that it will change soon and we can add things like:
```
app
   \_ commands
     \_ add_waveform.rb
     \_ ensure_unique_permalink.rb
     \_ notify_followers.rb
```

innatewonderer:
Note, can't seem to use `module Asset`
```
 TypeError:
        Asset is not a module
      # ./app/commands/asset/soft_delete_with_relations.rb:1:in `<top (required)>'
```

app/controllers/admin/assets_controller.rb
innatewonderer: We have a clear separation of `Spam/Unspam` from `Delete/Restore` in users.
I would prefer to keep those boundaries clear here as well. We might not want to delete spam tracks right away (for research purposes :). Especially if we have an automatic permanent deletion of an asset after a certain time. 

We could add a job that would `DeleteAllSpamAssets` or something very precise like that. Thoughts?

sudara: > We have a clear separation of Spam/Unspam from Delete/Restore in users.

Marking a user as a spam always soft deletes it: https://github.com/sudara/alonetone/blob/master/app/controllers/admin/users_controller.rb#L19-L24

I'm not sure we'd want to change that deviate from "spam means deleted" mental model without a good reason. The implication on the UI side is "users are spammed and deleted with one click but tracks require 2 clicks" — Is the reason technical? (i.e. that we'd have to switch out a lot of `not_spam` scopes for asset?)

sudara:
Ahh right, `Asset` is defined as a class (model). 

Some of my clients name commands explicitly like so: 

```
class AssetCommand
  def initialize(asset)
    @asset = asset
  end

  def delete_with_relations
    @asset.comments&.with_deleted&.delete_all
    # ...
  end

  def restore_with_relations
    # ...
  end

  def really_destroy_with_relations

  end
end
```

That makes for a friendly and direct API: `AssetCommand.new(@asset).delete_with_relations`

I'm not really sure we will have any more asset related commands coming up in addition to these (waveform can be `WaveformCommand` otherwise in an `AudioFeature` namespace) — but if you wanna namespace my vote is "anything but Assets" as that feels unintuitive (operates on 1 asset vs. many assets.)

Another kinda interesting theme — we are finding out that both soft and regular deletion both reliably have "after" callbacks....

sudara:
Followup: https://github.com/sudara/alonetone/issues/672

app/controllers/admin/assets_controller.rb
innatewonderer: Moved to #672

PR #654, opened 2019-08-13
Clone the play animation SVG before animating it.

### LINKED ISSUE #647: After navigating with paginator on user/tracks getting JS play errors
Error initializing controller

Error invoking action "click->normal-playback#togglePlay"

sudara:
We are still seeing glitching of the play animation svgs on playback.

Part of the reason why is because we have been working with just 1 svg vs. many tracks — so the state of the svg has to be manually managed and reset, etc. 

We revisited that decision, and now print a copy of the svg animation (`cloneNode`) to use each time play is clicked on a track. 

This seems to have resolved a lot of the problems.

Fixes #647


sudara:
image

Still seeing some issues. I'll make `play_animation.js` work on a local copy of the animation instead of a static id and see if we can just bypass the trouble.

PR #659, opened 2019-08-13
Refine big player design with samo playhead

### LINKED ISSUE #643: Saggy waveform
We should vertically center it — it'll help with it lining up better with the play button too.

Messages Image(164926213)

ofsound:
Fixes #643

@sudara 

Done fiddling with this, maybe a good thing for samo to check out as it is our ‘flagship’ player

But perhaps this state is more desirable than the time under the play button state that currently exists.

That's up to you!


sudara:
Fresh off the presses from @smoofles (also in figma)

Screen Shot 2019-08-16 at 14 38 39

sudara:
I'd like to get Percy rendering the waveform so we can better test playback — most ideally we'd move the waveform to svg (away from canvas). But we could also use this hack to render the canvas as an image when in the percy context https://gist.github.com/Robdel12/d4e42b8d7dd73a145ecde238454df764

sudara:
@ofsound Looking sweet! Does it make sense to you that we only show the playhead when playback has started?

ofsound:
> @ofsound Looking sweet! Does it make sense to you that we only show the playhead when playback has started?

Yeah, that would be my inclination, have the new orange elements only show once they would start moving.

As far as animating this, I'd guess something hit from whilePlayingCallBack() in single_playback_controller?

Would .line-progress and .time-progress be wrapped by an element that is its own Stimulus class?

ofsound:
> I'd like to get Percy rendering the waveform so we can better test playback — most ideally we'd move the waveform to svg (away from canvas). But we could also use this hack to render the canvas as an image when in the percy context https://gist.github.com/Robdel12/d4e42b8d7dd73a145ecde238454df764

An SVG waveform would be awesome! Seems pretty do-able.

sudara:
> As far as animating this, I'd guess something hit from whilePlayingCallBack() in single_playback_controller?

Exactly, although it's a bit awkward because we need to update both `single_playback_controller` and `playlist_playback_controller` — this is why `big_play_controller` exists — to be the place where anything related to the waveform and big button can happen.

Right now there is duplication in the `whilePlayingCallback()` that is ripe for moving into a method on `bigPlay` — then in addition to updating the waveform we can update the playhead.

```javascript
    this.bigPlay.percentPlayed = this.percentPlayed()
    this.bigPlay.waveform.update()
    this.bigPlay.timeTarget.innerHTML = this.time
```

> Would .line-progress and .time-progress be wrapped by an element that is its own Stimulus class?

Yeah, if they can be easily positioned together, that means you'll only need to add and update 1 stimulus target. Otherwise you could do them both. Do you think we'll just manually update position of the element?

sudara:
screencast 2019-08-30 03-18-00

app/javascript/controllers/single_playback_controller.js
```diff
@@ -23,10 +23,12 @@ export default class extends PlaybackController {
     if (!this.loaded) {
       this.bigPlay.animation.showPause()
       this.loaded = true
+      this.bigPlay.progressContainerInnerTarget.classList.add('visible')
     }
     this.bigPlay.percentPlayed = this.percentPlayed()
     this.bigPlay.waveform.update()
     this.bigPlay.timeTarget.innerHTML = this.time
+    this.bigPlay.progressContainerInnerTarget.style.left = 100 * this.percentPlayed() + "%"
```
sudara: @ofsound Since these 4 lines in this `if` condition will have to also be in `playlist_playback_controller.js` I was thinking we could instead call `this.bigPlay.update(this.percentPlayed())`

The `update` method could look something like this:

```javascript
update(percentPlayed) {
  this.percentPlayed = percentPlayed
  waveform.update
  this.timeTarget.innerHTML = this.delegate.time
  this.progressContainerInnerTarget.style.left = 100 * this.percentPlayed() + "%"
}
```

ofsound:
@sudara Please review my stimulus code from these last commits at your convenience

sudara:
@ofsound Looking good! Welcome to the team. #jstune #juniordev #stimulus #framework 

The codeclimate complaints are legit (more duplicated code that can be shared) but for now let's merge this! I'm not 100% it was the right move for setup the whole bigPlay/delegate abstraction given that it's only 2 places and the stimulus API to talk from controller to controller is a bit limited/hacky (`this.application.getControllerForElementAndIdentifier`)

PR #670, opened 2019-08-15
Remove hardcoded 2018 date in playlist edit and empty <li>s in external links

ofsound:
@sudara I tried to rubytune both of these changes myself, so let me know if there's a better solution. They both appeared to fix the issues.

* [ ] PR description
  * Does anything special need to happen for deployment?
* [ ] Optional: Inline github PR comments explaining context of tricky / complicated / unexpected lines so reviewers can follow along.
* [ ] If appropriate, new tests were added for isolated methods or new endpoints
* [ ] All tests green
* [ ] I booted up the branch locally, exercised any new code I wrote.
* [ ] If css was touched, I verified changes are happy on mobile (via Percy is ok)

### DIFF
--- a/app/views/playlists/_playlist_white.html.erb
+++ b/app/views/playlists/_playlist_white.html.erb
@@ -49,9 +49,11 @@
     <ul class="external-links">
       <% [:link1, :link2, :link3].collect do |link| %>
         <% if valid_link = @playlist.send(link) %>
+          <% if external_link_for(valid_link) %>
           <li>
             <%= external_link_for(valid_link) %>
           </li>
+          <% end %>  
         <% end %>
       <% end %>
     </ul>
--- a/app/views/playlists/edit_white.html.erb
+++ b/app/views/playlists/edit_white.html.erb
@@ -71,7 +71,7 @@
     </p>
     <div class="box">
       <div class="left_column_box_header">
-        <h2><%= @playlist.title %> <span class="playlist_date">(<%= @playlist.year %>)</span></h2>
+        <h2><%= @playlist.title %></h2>
         <div>
            <span class="playlist_track_size" data-target="playlist-sort.trackCount"><%= pluralize @playlist.tracks_count, 'track' %></span> <span class="playlist_dash">&mdash;</span>
            <span class="playlist_total_time" data-target="playlist-sort.totalTime"><%= @playlist.play_time %></span>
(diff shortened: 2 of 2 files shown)


app/views/playlists/_playlist_white.html.erb
```diff
@@ -49,9 +49,11 @@
     <ul class="external-links">
       <% [:link1, :link2, :link3].collect do |link| %>
         <% if valid_link = @playlist.send(link) %>
+          <% if external_link_for(valid_link) %>
```
sudara: Thanks for this, it will work for now. It will have to be refactored eventually – `external_link_for` is checking to see if the link exists, but what we want is to have that check happen before the `<li>` is written. That means we should either put the check for presence outside the `li` or put the `li` inside the helper (these days i don't like hiding html in helpers, really...)

app/views/playlists/edit_white.html.erb
```diff
@@ -71,7 +71,7 @@
     </p>
     <div class="box">
       <div class="left_column_box_header">
-        <h2><%= @playlist.title %> <span class="playlist_date">(2018)</span></h2>
+        <h2><%= @playlist.title %> <span class="playlist_date">(<%= @playlist.year %>)</span></h2>
```
sudara: Percy is showing that this opens up a problem when a date isn't present — we get empty parens...We probably want to wrap the whole span in an `if playlist.year.present?`

ofsound: @sudara 

I might argue against having the date here at all, as horizontal space is a premium here for long playlist titles, and the date is clearly in the form at the top of the page?

sudara: Even better — lets kill it!

sudara:
`if valid_link = @playlist.send(link)` is doing something a bit too tricky. 

The `=` here isn't a conditional, it's assignment. So `valid_link` is getting the value of `@playlist.send(link)` — this means we can use `valid_link` later....

`@playlist.send(link)` is taking the value of `link` and calling it as a method on `@playlist`. So in the case of `link1` that piece of code translates to calling `@playlist.link1`

In the end `if @playlist.link1` is implicitly asking "is there a value here?". The problem is "" is considered a value, it's just an empty value. So really it needs to be asking  `if @playlist.link1.present?` 

Basically, your if condition ends up needed because there's a bug in the fancy footwork.

PR #671, opened 2019-08-15
Attempt to debug mysterious soft-deleted user.

### LINKED ISSUE #653: Assets soft deleted from a user's soft delete are not handling all dependencies (track, comment(?), etc)
In the case of this bugsnag, the `asset` is present, but deleted (probably because its parent user was deleted) — so in the context of a playlist, the `track.asset.id` is `nil`.

```
irb(main):011:0> playlist.tracks.collect {|t| Asset.with_deleted.find(t.asset_id) }.count
=> 69
irb(main):012:0> playlist.tracks.collect {|t| Asset.find(t.asset_id) }.count
Traceback (most recent call last):
        3: from (irb):12
        2: from (irb):12:in `collect'
        1: from (irb):12:in `block in irb_binding'
ActiveRecord::RecordNotFound (Couldn't find Asset with 'id'=34775 [WHERE `assets`.`deleted_at` IS NULL])
irb(main):014:0> Asset.with_deleted.find(34775)
=> #<Asset id: 34775, mp3_content_type: "audio/mpeg", mp3_file_name: "stronger_together.mp3", mp3_file_size: 8845695, created_at: "2011-03-12 22:11:27", title: "Stronger Together", thumbnails_count: 0, user_id: 4354, length: 221, album: nil, permalink: "stronger-together", samplerate: 48000, bitrate: 320, genre: nil, artist: "Dave McCormick", listens_count: 198, description: "", credits: nil, youtube_embed: "", hotness: 0.00212567, favorites_count: 2, lyrics: nil, description_html: nil, listens_per_week: 0.0, comments_count: 4, updated_at: "2019-03-03 01:33:26", private: false, id3_track_num: 1, is_spam: false, deleted_at: "2019-04-04 08:08:30", user_agent: nil>
```

## Error in alonetone

**ActionView::Template::Error** in **playlists#show**
undefined method `id' for nil:NilClass

[View on Bugsnag](https://app.bugsnag.com/alonetone/alonetone/errors/5d11938673d2380018dee246?event_id=5d4fa727004ccce11ae50000&i=gh&m=ci)

## Stacktrace

    app/views/playlists/_playlist_white.html.erb:14 - block in _app_views_playlists__playlist_white_html_erb__321828785332572429_69857178009000
    app/views/playlists/_playlist_white.html.erb:13 - _app_views_playlists__playlist_white_html_erb__321828785332572429_69857178009000
    app/views/playlists/show_white.html.erb:8 - _app_views_playlists_show_white_html_erb__2439390540724164419_69857177562100
    app/controllers/playlists_controller.rb:48 - block (2 levels) in show
    app/controllers/playlists_controller.rb:41 - show

[View full stacktrace](https://app.bugsnag.com/alonetone/alonetone/errors/5d11938673d2380018dee246?event_id=5d4fa727004ccce11ae50000&i=gh&m=ci)

*Created by sudara via Bugsnag*

innatewonderer:
Investigation into #653 

-  including a spec in this PR that confirms that all tracks are indeed deleted when a user and its assets are soft_deleted. 

Things I tried that didn't reproduce the bug:
:x: Soft-delete an `asset`. Observed all `tracks` are deleted
:x: Soft-delete a `user`. Observe all `assets` and `tracks` are deleted
:x: Investigated into `asset has_many :tracks` vs ` asset has_many :favorite tracks`. Both were deleted if `asset` is deleted
:x: Private playlist didn't make a difference in the above scenarios
:x: SoftDeleted user who is an asset owner, vs use who is an owner of the track. Both got deleted as part of `efficiently_delete_relations`
```
# user owned assets
Track.joins(:asset).where(assets: { user_id: user.id }).update_all(deleted_at: time)
# vs user's owned tracks
%w[tracks playlists comments].each do |user_relation|
    user.send(user_relation).update_all(deleted_at: time)
end
```
My current hypotheses are that if somehow we soft-deleted an asset (in the console) without going through the proper commands that take care of all relations then it would create this particular situation. 
```
2.6.0 :030 > Track.last
  Track Load (1.2ms)  SELECT `tracks`.* FROM `tracks` WHERE `tracks`.`deleted_at` IS NULL ORDER BY `tracks`.`id` DESC LIMIT 1
 => #<Track id: 151597, playlist_id: 10472, asset_id: 63602, position: 13, created_at: "2019-04-13 19:53:28", updated_at: "2019-04-13 19:53:28", is_favorite: true, user_id: 12880, deleted_at: nil>
2.6.0 :031 > track = _
 => #<Track id: 151597, playlist_id: 10472, asset_id: 63602, position: 13, created_at: "2019-04-13 19:53:28", updated_at: "2019-04-13 19:53:28", is_favorite: true, user_id: 12880, deleted_at: nil>
 2.6.0 :035 > track.asset.soft_delete
   (0.3ms)  BEGIN
  Asset Update (1.3ms)  UPDATE `assets` SET `assets`.`updated_at` = '2019-08-18 16:43:44', `assets`.`deleted_at` = '2019-08-18 16:43:44' WHERE `assets`.`id` = 63602
   (5.1ms)  COMMIT
 => true
2.6.0 :037 > track.reload
  Track Load (1.2ms)  SELECT `tracks`.* FROM `tracks` WHERE `tracks`.`id` = 151597 LIMIT 1
 => #<Track id: 151597, playlist_id: 10472, asset_id: 63602, position: 13, created_at: "2019-04-13 19:53:28", updated_at: "2019-04-13 19:53:28", is_favorite: true, user_id: 12880, deleted_at: nil>
2.6.0 :038 > track.asset
  Asset Load (1.6ms)  SELECT `assets`.* FROM `assets` WHERE `assets`.`deleted_at` IS NULL AND `assets`.`id` = 63602 LIMIT 1
 => nil
2.6.0 :039 > playlist.tracks.collect {|t| Asset.with_deleted.find(t.asset_id) }.count
  ...
  ...
 => 13
2.6.0 :040 > playlist.tracks.collect {|t| Asset.find(t.asset_id) }.count
...
Traceback (most recent call last):
        3: from (irb):40
        2: from (irb):40:in `collect'
        1: from (irb):40:in `block in irb_binding'
ActiveRecord::RecordNotFound (Couldn't find Asset with 'id'=63602 [WHERE `assets`.`deleted_at` IS NULL])
```
Still looking.

### DIFF
--- a/app/commands/user_command.rb
+++ b/app/commands/user_command.rb
@@ -1,0 +1,72 @@
+class UserCommand
+  attr_reader :user
+  def initialize(user)
+    @user = user
+  end
+
+  def soft_delete_with_relations
+    efficiently_delete_relations
+    user.soft_delete
+  end
+
+  def restore_with_relations
+    efficiently_restore_relations
+    user.restore
+  end
+
+  def destroy_with_relations
+    efficiently_destroy_relations
+  end
+
+  private
+
+  def efficiently_delete_relations(time = Time.now)
+    Listen.where(track_owner_id: user.id).update_all(deleted_at: time)
+    Listen.where(listener_id: user.id).update_all(deleted_at: time)
+    Topic.where(user_id: user.id).where('posts_count < 2').destroy_all
+    Playlist.joins(:assets).where(assets: { user_id: user.id })
+            .update_all(['tracks_count = tracks_count - 1, playlists.updated_at = ?', time])
+    Track.joins(:asset).where(assets: { user_id: user.id }).update_all(deleted_at: time)
+    Comment.joins("INNER JOIN assets ON commentable_type = 'Asset' AND commentable_id = assets.id")
+           .joins('INNER JOIN users ON assets.user_id = users.id').where('users.id = ?', user.id).update_all(deleted_at: time)
+
+    user.assets.update_all(deleted_at: time)
+
+    %w[tracks playlists comments].each do |user_relation|
+      user.send(user_relation).update_all(deleted_at: time)
+    end
+  end
+
+  def efficiently_restore_relations
+    Asset.with_deleted.where(user_id: user.id).update_all(deleted_at: nil)
+    Listen.with_deleted.where(track_owner_id: user.id).update_all(deleted_at: nil)
+    Listen.with_deleted.where(listener_id: user.id).update_all(deleted_at: nil)
+    Playlist.with_deleted.joins(:assets).where(assets: { user_id: user.id })
+            .update_all(['tracks_count = tracks_count - 1, playlists.updated_at = ?', Time.now])
+    Track.with_deleted.joins(:asset).where(assets: { user_id: user.id }).update_all(deleted_at: nil)
+    Comment.with_deleted.joins("INNER JOIN assets ON commentable_type = 'Asset' AND commentable_id = assets.id")
+           .joins('INNER JOIN users ON assets.user_id = users.id').where('users.id = ?', user.id).update_all(deleted_at: nil)
+
+    Track.with_deleted.where(user_id: user.id).update_all(deleted_at: nil)
+    Playlist.with_deleted.where(user_id: user.id).update_all(deleted_at: nil)
+    Comment.with_deleted.where(user_id: user.id).update_all(deleted_at: nil)
+  end
+
+  def efficiently_destroy_relations
+    Listen.where(track_owner_id: user.id).delete_all
+    Listen.where(listener_id: user.id).delete_all
+    Playlist.joins(:assets).where(assets: { user_id: user.id })
+            .update_all(['tracks_count = tracks_count - 1, playlists.updated_at = ?', Time.now])
+    Track.joins(:asset).where(assets: { user_id: user.id }).delete_all
+
+    Comment.joins("INNER JOIN assets ON commentable_type = 'Asset' AND commentable_id = assets.id")
+           .joins('INNER JOIN users ON assets.user_id = users.id').where('users.id = ?', user.id).delete_all
+
+    user.assets.destroy_all
+
+    %w[tracks playlists posts comments].each do |user_relation|
+      user.send(user_relation).delete_all
+    end
+    true
+  end
+end
--- a/app/commands/user_command.rb
+++ b/app/commands/user_command.rb
@@ -18,6 +18,18 @@ class UserCommand
     efficiently_destroy_relations
   end
 
+  def spam_soft_delete_with_relations
+    user.spam!
+    user.update_attribute :is_spam, true
+    soft_delete_with_relations
+  end
+
+  def unspam_and_restore_with_relations
+    user.ham!
+    user.update_attribute :is_spam, false
+    restore_with_relations
+  end
+
   private
 
   def efficiently_delete_relations(time = Time.now)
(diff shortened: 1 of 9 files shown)


app/controllers/admin/users_controller.rb
```diff
@@ -7,24 +7,26 @@ def index
...
     def unspam
       @user.ham!
       @user.update_attribute :is_spam, false
-      @user.restore
-      @user.restore_relations
+      UserCommand.new(@user).restore_with_relations
     end
 
     def spam
-      @user.spam_and_mark_for_deletion!
```
sudara: We should find a way to keep the "spam and delete" logic encapsulated by 1 method. There are multiple places that call `spam_and_mark_for_deletion!` and I use it on console. It was made because we don't want a visible user marked as spam in the UI.

spec/fixtures/tracks.yml
```diff
@@ -101,3 +101,10 @@ soft_deleted_track_on_asset:
   asset: asset_with_relations_for_soft_delete
   position: 1
   created_at: <%= 2.days.ago %>
+
+track_on_soft_deleted_of_other_user:
```
sudara: This track fixture isn't yet deleted, which I think makes this name difficult to parse. Maybe something like `track_with_asset_from_other_user` or `sudaras_track_from_arthurs_asset` hmmmm...

spec/request/admin/users_controller_spec.rb
```diff
@@ -111,6 +111,17 @@
...
+
+    it "soft deletes from other user's playlist" do
+      # sudara's playlist
+      playlists(:owp)
+      # arthur's track on owp playlist
+      tracks(:track_on_soft_deleted_of_other_user)
+
+      put delete_admin_user_path(users(:arthur))
+
+      expect(tracks(:track_on_soft_deleted_of_other_user).reload.deleted_at).not_to be_nil
```
sudara: [Change Matcher](https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/built-in-matchers/change-matcher) might be your friend and prevent the need for reload call:

expect { put delete_admin_user_path(users(:arthur)) }.to change(track.deleted_at)

app/commands/user_command.rb
```diff
@@ -0,0 +1,72 @@
...
+    user.soft_delete
+  end
+
+  def restore_with_relations
+    efficiently_restore_relations
+    user.restore
+  end
+
+  def destroy_with_relations
+    efficiently_destroy_relations
```
sudara: Zen Koan: Is there value to a method that only calls another private method?

sudara:
Because the logic moved, it's for me to see what was added removed with regards to the primary purpose of the PR — did you add a query to soft deletion?

> My suspicion for the bugsnag bug is that the URL was being hit directly by a bot for an asset that was indeed deleted (an entire playlist was indeed deleted here)

I'm assuming you mean entire *user*, though then I don't follow the bot thing — From what I'm seeing, the playlist is publicly available, but rendering the view errors because one of the track's asset is `nil`.

innatewonderer:
Another thing I tried that didn't work:
:x: Tried deleting an album playlist. Was wondering if maybe deleting an album, would delete assets but skip other tracks. But we actually aren't deleting `assets` when `playlist` is deleted.

sudara:
Nice work! I've closed #653 on the assumption it's a one off data integrity issue from some console work. Bugsnag will yell if there's still a problem.

Feel free to reframe this PR in context of the commands work or close it out to organize however you like!

Issue #672, opened 2019-08-15
Soft Delete Tracks marked as spam


sudara:
Followup to #636.

Assets marked as spam currently do *not* show in the UI because we manually pepper in the the scope `not_spam` everywhere.

We want to keep that UI behavior but move to the more formal mechanism to remove tracks from the UI (soft deletion) — This is in line with how user spam+soft deletion works.

* [x] Everywhere an asset is marked as spam, we need to soft delete
* [x] Remove `not_spam` scope
* [x] Make sure admin UI is sane with buttons — should be similar to users UI in terms of what buttons show

innatewonderer:
[ ] include one-shot (migration) to soft_delete all spam assets (otherwise they will all show up on Latest.

innatewonderer:
<img width="767" alt="Soft_Delete_Tracks_marked_as_spam_·_Issue__672_·_sudara_alonetone" src="https://user-images.githubusercontent.com/2320322/63198370-80a53600-c048-11e9-81b4-5ad0e1515af9.png">

sudara:
Fixed by #673

PR #673, opened 2019-08-16
I672 spam delete assets

innatewonderer:

* Issues covered in this PR:
  - #672 
* Changes:
  - Move asset spam and soft delete functionality into its own [method](https://github.com/sudara/alonetone/pull/673/files#diff-c8937128ca0257e1aaa93a93ce0adce7R35-R40) with supporting specs
  - Remove `not_spam` scope (no specs failed)
  - Remove `is_spam: false` from `published` scope (with supporting spec)
  - Add migration to soft_delete all spam assets
  - Cleanup unused `mark_group_as_spam` [method](https://github.com/sudara/alonetone/pull/673/files#diff-8845870d4db7fd765bff8706527cfc92L24) from `admin/assets_controller`
  - Add `filter_by` method to Asset to handle how admin/assets is displayed:
    - is_spam displays `deleted spam` records
    - not_spam displays `not_deleted not_spam` records
    - deleted displays `deleted not_spam` records
  - Add proper redirects when we Spam, Delete or Restore an asset:
    - Spam button will redirect to Deleted (Spam)
    - Delete button will redirect to Delete (Not Spam)
    - Restore redirects to Not Spam 
  
- Cleanup admin/assets Stimulsjs controller:
    - since Spam button is similar in behaviour as Delete button, we will trigger the same javascript behaviour for both.
    - refactor following admin/users stimulus controller pattern

* Implementation Details:

`not_spam` scope doesn't appear to be used anywhere except an unused method in `admin/assets_controller`. It is `is_spam: false` flag in `scope :published` that seem to prevent the assets from being displayed on `/latest` page. 

- [x] Including inline comments
- [x] Including new tests
- [x] Including some new fixtures. Percy complained, I confirmed that it's related to the two fixtures I added. I approved Percy changes.
- [x] I booted up the branch locally, exercised any new code I wrote.

Merged PR: #746


sudara:
After migrating locally I got this error:

Action Controller: Exception caught 2019-09-03 20-56-07

sudara:
The error only happens on "all tracks" — maybe the user relation needs to be found `.with_deleted`?

Also, I'm seeing 782 tracks under Deleted (not spam) but none of them display in the UI

image

innatewonderer:
I added `with_user` to navigation counter (since we only display assets that were deleted on its own rather than through its user. 
I am not seeing the same error on /index page (my only snapshot is of 2019.04)... wonder if it's an issue with data? Is it possible your snapshot contains that mysterious record that was causing our Bugsnag issue?

app/models/asset.rb
```diff
@@ -29,6 +28,8 @@ class Asset < ApplicationRecord
   scope :hottest,         -> { where('hotness > 0').order('hotness DESC') }
   scope :most_commented,  -> { where('comments_count > 0').order('comments_count DESC') }
   scope :most_listened,   -> { where('listens_count > 0').order('listens_count DESC') }
+  scope :with_user,       -> { joins(:user).where('users.id IS NOT NULL') }
```
sudara: I believe this is a workaround because we are having trouble "reaching" users from an asset when they are deleted. I think the answer is to unscope the default user scope on `belongs_to :user` below, adding `.with_deleted` — I'm not 100% sure how that affects the counter cache.....I guess we want the `assets_count` to decrement when a track is deleted — maybe we have specs in place for this...

sudara:
I think the error on index will go away when the scope is in place. You can test this by soft deleting a user that has assets and seeing if admin/assets/ loads correctly.

Following up from slack, we talked about still showing assets, even when their user is deleted, but disallowing restoration: https://alonetone.slack.com/archives/DC8SK0KRD/p1566137045026500

In that case, we could ideally have "show deleted user" as the action instead of "restore", which would link to the user admin. (we can use email address as filter_by to link to the 1 user in the admin).

innatewonderer:
Bug: #753

PR #674, opened 2019-08-18
Update PR template to encourage prose, detail, rational.

sudara:
### What changes are in this PR? Why did you implement them this way?

This PR adjusts the PR template to move away from a "checklist-only" description by 

1) Adding a few prose prompts
2) Adding a prompt for bug reproduction  
3) Reducing the length of the checklist and splitting it into 2 parts
4) Asking PR author to remove the checklists to reduce noise

We want to upgrade our PR descriptions from "a list of things the code does" to more of a first class representative of the changes taking place in the PR. 

Longer descriptions with the rationale and process included encourages deeper thinking. Writing them typically helps bubble up additional concerns that might otherwise not come up until code review. It also communicates more context to the reviewer.

### Was anything tried that didn't work? Is there anything reviewers need to pay attention to? Anything difficult or tricky that should be explained?

I looked into how we could create 2 different templates: bug and feature — that looks possible, but you can only actually switch between them by passing the `template` url variable and not via the UI (this is possible with issue templates, however). So for now I included the prose prompt for bug reproduction up top and it will have to be manually deleted when not applicable to the PR.

Another open issue is "noise" — for example, on this PR, almost none of the checklist is relevant. The "bug reproduction" prompt is also not relevant. To increase readability, I'd like to manually curate what is put in the PR description vs. including it all by default. 

Another concern is part of the checklist should actually happen after every commit (tests/percy/etc). It's somewhat confusing to see PRs with ✅ that tests passed, and then a red X on the actual tests. We should let the tests/percy/etc speak for themselves and have the checklist be a "reminder" that gets removed when a PR is opened.

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,17 +1,22 @@
 ### [Bug] Provide step by step console or UI reproduction steps. When and how did the problem begin?
-### What changes are in this PR and why did you implement it this way?
 
-### Is there anything reviewers need to pay attention to? Anything difficult or tricky that should be explained?
+### What changes are in this PR? Why did you implement them this way?
+
+### Was anything tried that didn't work? Anything that reviewers should pay attention to or difficult or tricky that should be explained?
 
 ### Does anything special need to happen for deployment?
 
 ## "Ready For Review" checklist
 
+* [ ] Fixes #...
 * [ ] PR title accurately summarizes changes
-* [ ] Optional: Inline github PR comments explaining context of tricky / complicated / unexpected lines so reviewers can follow along.
-* [ ] If appropriate, new tests were added for isolated methods or new endpoints
-* [ ] All tests green
-* [ ] Percy changes are purposeful or explained
-* [ ] I booted up the branch locally, exercised any new code I wrote
-* [ ] If css was touched, I verified changes are happy on mobile (via Percy is ok)
+* [ ] New tests were added for isolated methods or new endpoints
 * [ ] I opened an issue for any logical followups
+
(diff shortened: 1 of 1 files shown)


pull_request_template.md
```diff
@@ -1,18 +1,29 @@
+### [Bug] Step by step walkthrough of how to reproduce (console/UI/etc). When and how did the problem begin?
+
+
+### Iterate through the changes in this PR. Why did you implement them this way?
+
+
+### Was anything tried that didn't work? Anything that reviewers should pay attention to or difficult or tricky that should be explained?
```
innatewonderer: `Was anything tried that didn't work?`
Yes 👍 This is awesome.

innatewonderer:
What do you think about adding something related to an Issue a PR might be fixing? @sudara, you often put a lot of thought in writing up the issue and its proposed solution. Feels like we should utilize it somehow.
Should we: 
- just link the issue and expect the reviewers to read it
- briefly summarize the issue and the root cause/solution inside the PR description
- copy/paste pieces from the issue.

sudara:
My preference is that we don't have to leave the PR to get full context. If the PR closes out an issue, that should be listed first/last in the description.

In most cases, after the work has happened (PR) there's a different perspective than before happens (issue). The scope shifts, new things bubble up, some things are reprioritized, etc. 

I like to use issues as "source material" and listing out work before the it happens — they are  useful to refer to/crib from, but don't replace framing what actually happened (and why) in a PR.

PR #675, opened 2019-08-18
Bump to Rails 6 final

sudara:
This was the result of `bundle update` after removing version specificity on Rails and webpacker.

`sass-rails` and `compass` both jumped major versions and need due diligence.


Gemfile.lock
```diff
@@ -148,10 +148,8 @@ GEM
       sass (>= 3.3.0, < 3.5)
     compass-import-once (1.0.5)
       sass (>= 3.2, < 3.5)
-    compass-rails (3.1.0)
+    compass-rails (2.0.1)
```
sudara: Hmmmmmmmmmmmmmmm.

Gemfile.lock
```diff
@@ -466,7 +460,7 @@ GEM
       climate_control (>= 0.0.3, < 1.0)
     thor (0.20.3)
     thread_safe (0.3.6)
-    thredded (0.16.12)
+    thredded (0.16.13)
```
sudara: `.13` is the only release compatible with rails 6.0 final.

sudara:
Looks like we can't upgrade `sass-rails` until we remove dark theme and compass https://github.com/sass/sassc-rails/issues/79

PR #683, opened 2019-08-20
Bjm admin css tweaks

### LINKED ISSUE #619: Admin css tweaks
1. Search box might be nice to float right( in horizontal line with pagination). Or otherwise style as you like!

2. I'd like to show the "deleted at" date but have it be a different color.
image

3. Would be sweet if hovers went full width of nav.
image

4. I'd like to add counts to the navigation like "Only Spam (34)", would be cool to get a placeholder there.

sudara:
Closes #619


sudara:
@ofsound Can we tweak the counts to compete less with the navigation item name? Maybe gray and/or floating right.

image

sudara:
Nice!!!

PR #709, opened 2019-08-30
Refactor user soft-delete actions to follow Command Pattern

innatewonderer:
- This PR just moves all soft-related user functionality to app/commands/user_command. For consistency with `assets`
- It also contains a spec I added to confirm that all `tracks` (not only own tracks but also tracks of other users on this one asset) are deleted if a user is soft-deleted.


app/commands/user_command.rb
```diff
@@ -0,0 +1,84 @@
...
+    user.soft_delete
+  end
+
+  def restore_with_relations
+    efficiently_restore_relations
+    user.restore
+  end
+
+  def destroy_with_relations
+    efficiently_destroy_relations
```
innatewonderer: I know you questioned having a method that only points to another method - but this one hides the actual implementation to private method. OK to leave it as is?

sudara:
Feels great to move this group of stuff out of `user.rb` — thanks for the relocation. It might be nice to add some specs around the public methods of the command (in general I prefer that if we are adding classes, we also add specs) but I'm going to merge this for now for the sake of moving things forward!

Issue #696, opened 2019-08-23
Clean up css color names in prep for dark theme


sudara:
Samo is codifying and naming colors in figma, such as `red500` which is darker than `red200`. Goal:  improve consistency/clarity/reusability & enable theming. 

We chatted about moving css variables to more semantic names — instead of `$samo-background-orange` we move to more of a semantic naming setup like `$primary-accent-color`.

As colors get used in more and more locations, there isn't a great way to pick 1 name that accurate describes each location. Two options to handle: 

1) Suck it up and elect a representative by popular vote — i.e. be happy using `$main-text-color` in many other places and have that be a bit of insider knowledge you have to get comfy with.

2) Go granular/component-y and add another layer of abstraction. 

```css 
// color-definitions.css
$gray500: #353535;
$gray200: #868686;

// white-theme.css
$save-button-border: $gray500;

// dark-theme.css
$save-button-border: $gray200;

// buttons.css
border-color: $save-button-border;
```

One side effect of this route is that the theme files become a bit bloated and verbose, as everything and its mother has a css variable name.

Regardless of the option we choose, at bare minimum we need to remove the "curse of the ambiguous grays" — we should aim to align these css variable names with samo's new names in figma.

<img width="408" alt="image" src="https://user-images.githubusercontent.com/472/63597680-53173a00-c5be-11e9-9cfa-ef80df9802bf.png">

sudara:
This is how I see things iterating:

* [x] @ofsound Consolidate hex values so css only uses variables
* [ ] Communicate colors & elements used on live site not represented in figma to @smoofles 
* [ ] @ofsound/@sudara choose semantic names (such as `$primary-text-color`) for existing colors used on the site
* [ ] @smoofles Finalizes WT color palette 
* [ ] @ofsound Imports color palette into new `white_theme_mapping.css` file 
* [ ] @ofsound makes adjustments to css (probably a reduction in number of variables) to map to palette

PR #712, opened 2019-09-02
Dark Theme Example

sudara:
* HTML will require `white_theme.scss` or `dark_theme.scss`
* Those two files will load up all the `components` as well as the relevant theme file and color mappings.

TODOS
* [ ] Grab all color names from figma and put them in `color_mappings.scss`
* [ ] Give semantic names to all existing color variables in `themes/white.scss`
* [ ] Map those fresh new semantic color variables to the named colors in `color_mappings.scss` — if we find some variables are using colors not in figma, check figma to see if they've been "adjusted" to one in the figma pallete. If the color isn't present in figma but seems needed, add it to this issue: https://github.com/sudara/alonetone/issues/696


sudara:
#757 did this in real life.

PR #717, opened 2019-09-03
Fixes #631 Add stimulus controller for header search icon

### LINKED ISSUE #631: Magnifying Glass Search icon in header doesn't seem hooked up, but hitting enter will initiate search

ofsound:
### [Bug] Step by step walkthrough of how to reproduce (console/UI/etc). When and how did the problem begin?

This hadn't been added since handlers.js and jQuery went away.

### Iterate through the changes in this PR. Why did you implement them this way?

I copied the style and implementation of the dropdown controller.

* [ ] All tests green

### DIFF
--- a/app/javascript/controllers/header_search_controller.js
+++ b/app/javascript/controllers/header_search_controller.js
@@ -1,0 +1,10 @@
+import { Controller } from 'stimulus'
+
+export default class extends Controller {
+  
+  submitForm() {
+    const searchForm = document.getElementById('search_form')
+    searchForm.submit();
+  }
+
+}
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/header_search_controller.js
```diff
@@ -0,0 +1,10 @@
+import { Controller } from 'stimulus'
+
+export default class extends Controller {
+  
+  submitForm() {
+    const searchForm = document.getElementById('search_form')
```
sudara: This is what stimulus `targets` are for. You can look at other controllers to pick up the pattern, but essentially you define any element you intend to interact with:

```
static targets = ['form']
```

And then in the html you annotate the element you want connected with `data-target="header-search.form"`

Then you can do `formTarget.submit()`

sudara: If the controller was on the form instead of the div, you could call `this.element.submit()` (you wouldn't need to go through the hassle of defining an extra target) — however, in this case the image is outside the form, so it makes sense to have the controller on the div IMO.

Issue #722, opened 2019-09-04
Style `'Upload a new version of this track` section 


innatewonderer:
Does this need a little love:
<img width="1039" alt="Fullscreen_9_3_19__9_15_PM" src="https://user-images.githubusercontent.com/2320322/64223655-c6fcf080-cea2-11e9-8a97-5a1872a9fade.png">

Fix to open that section is coming in #313

sudara:
Resolved by #723

PR #723, opened 2019-09-04
Fixes#313 Unable to replace track with new file on white_theme

### LINKED ISSUE #313: Ability to replace an mp3 in-place on white theme
This exists on the dark theme: 

alonetone a damn fine home for musicians 2018-11-24 00-54-21

But apparently it's broken:

inbox sudara modernthings net 2018-11-24 00-52-21

Todos:

* [ ] add a place to do this on the white theme track edit page
* [ ] figure out why it's broken for normal users
* [ ] add a spec

innatewonderer:
Resolves #313 
- Upload/replace functionality works on white-theme (confirmed both locally and in specs).
- The issue was that javascript was listening for the wrong class and needed an id handle to open the hidden pane. FIxed both the js' class name in `querySelectorAll('.slide_open_next')`  and added `#` to the view to be able to locate `id=` div in `document.querySelector(id)`.

I was accidentally able to reproduce the issue in dark theme but tinkering with `white_theme_enabled` in the application controller. But since we are not going to keep that around for much longer, I didn't dig in why it might be happening.

* [x] I opened an issue for any logical followups. #722
* [ ] All tests green


app/javascript/packs/application.js
```diff
@@ -42,7 +42,7 @@ function handlers() {
     }
   })
 
-  document.querySelectorAll('.slide_open_href').forEach((link) => {
+  document.querySelectorAll('.slide_open_next').forEach((link) => {
```
sudara: I believe profile editing currently uses this code with `.slide_open_href` and would break.

app/views/assets/_edit_white.html.erb
```diff
@@ -28,7 +28,7 @@
     <% if @allow_reupload %>
 
     <div class="track-edit-bottom">
-      <%= link_to 'Upload a new version of this track',"new_version_#{asset.object_id}", class: 'slide_open_next'%>
+      <%= link_to 'Upload a new version of this track',"#new_version_#{asset.object_id}", class: 'slide_open_next'%>
```
sudara:
```suggestion
      <%= link_to 'Upload a new version of this track',"#new_version_#{asset.object_id}", class: 'slide_open_href'%>
```

This would solve the issue without breaking profile editing, I think.

spec/controllers/assets_controller_spec.rb
```diff
@@ -88,14 +87,67 @@
...
+
+      let(:asset) { users(:jamie_kiesl).assets.last }
+      subject {  put :update, params: { id: asset, user_id: users(:jamie_kiesl).login, asset: { mp3: fixture_file_upload('files/tag1.mp3', 'audio/mpeg') } } }
+
+      it 'should allow user to upload new version of song' do
+        post :create, params: { user_id: users(:jamie_kiesl).login, asset_data: [fixture_file_upload('files/muppets.mp3', 'audio/mpeg')] }
+        expect(users(:jamie_kiesl).assets.first.mp3_file_name).to eq('muppets.mp3')
+        put :update, params: { id: users(:jamie_kiesl).assets.first, user_id: users(:jamie_kiesl).login, asset: { mp3: fixture_file_upload('files/tag1.mp3', 'audio/mpeg') } }
+        expect(users(:jamie_kiesl).assets.reload.first.mp3_file_name).to eq('tag1.mp3')
+      end
```
sudara: 🥇

app/views/assets/_edit_white.html.erb
innatewonderer: Good idea. I looked over current usage, and for some reason decided that it's not used... I'll fix it.

sudara:
Percy was my fault — I changed the timing on the "headway" widget to be 1 second, which makes it suddenly show up in the screenshots. I approved the changes.

sudara:
@ofsound Would you be willing to take some temporary measure to resolve #722 in this branch?

ofsound:
> @ofsound Would you be willing to take some temporary measure to resolve #722 in this branch? 💄 🐷

@sudara

I took some general styles from the upload a new album cover section.

Does the new file get uploaded on save, if so should that button change its label? Maybe a Upload File button would be friendlier?

I've also wondered about a "Cancel Button" on this template, what if you don't want to save your changes?

Screenshot 2019-09-06 19 28 35

sudara:
@ofsound nice! 

Good point about the button UX.  Agreed that it should be “Upload New Version” or “Upload and Replace” or something more explicit. We could move the “slide open” pattern to be done in stimulus and add the label change. 

Feels like with playlist/edit and now this, it’s time for a little stimulus sprinkle that disables the save button until changes are made... 

Also, why is this save button hugging the left side, I wonder. Should it be on the right hand side to be consistent with our other forms?

sudara:
Hai Faifs all around, as @smoofles would say. Let's merge this and handle save button patterns in #733

PR #746, opened 2019-09-12
Unscope belongs_to user for asset to be able to display deleted user 

innatewonderer:
This PR unscopes the `asset belongs_to :user` relations to be able to access deleted user in `admin/assets` view.
I want to make sure that we do not display any assets who might have a user that was deleted by intention or an accident. In order to ensure that, I still have to add `with_user` scope to `scope :published` (which is our "it's OK to display with asset" scope).


innatewonderer:
Bug: #753

app/models/asset.rb
```diff
@@ -28,10 +28,14 @@ class Asset < ApplicationRecord
   scope :hottest,         -> { where('hotness > 0').order('hotness DESC') }
   scope :most_commented,  -> { where('comments_count > 0').order('comments_count DESC') }
   scope :most_listened,   -> { where('listens_count > 0').order('listens_count DESC') }
-  scope :with_user,       -> { joins(:user).where('users.id IS NOT NULL') }
   scope :recently_updated, -> { order('assets.updated_at DESC') }
 
   belongs_to :user, counter_cache: true
+  belongs_to :user_object,
```
sudara: `:user_object` is a bit confusing / doesn't help me understand why this other scope needs to exist. I would prefer something that more explicitly described why there's another relation, such as `:possibly_deleted_user` or `:with_deleted_user`....or.... #naming_is_hard 

A comment might also be nice — remind our future selves why this is needed (admin area only, etc).

innatewonderer: My thought was to separate `user` the scoped/active user from as close to whatever we have in a database (no filters or judgement :). So just an `object` of a user. Hence the name. If we stick with that, rather than something more specific like `possibly_deleted_user` we will never be able to reuse the name if we want to `unscope` it by some other parameter.
Some other words I considered:
```
- user_scema
- unscoped_user
- just_user
- frame_
- basic_
- structure_
```
I still think we should stick with something more neutral. Thoughts?

sudara: IMO the primary aim with naming methods/scopes/relations is to make sure it's understandable and intuitive in context. When another dev sees `asset.possibly_deleted_user` there's hopefully a bit of understanding about what that means and why it's being used instead of the more normal `asset.user`.  It's secondary for a name to work in hypothetical future cases — when that future comes and the meaning changes in context, it's then our job to refactor the name to fit. 

`object` has a very specific meaning in ruby, so it's likely to confuse devs — `Object` is the root of all objects and `.object_id` can be called on any object, including an AR object, etc. 

`asset.unscoped_user` might work, but it's ambiguous and technically confusing — the user query that fires is in fact "scoped" to the `asset_id`.

`asset.user_without_default_scope` could work, but it's chewier and doesn't actually "say what it means" — which is the user can be a deleted user.

`asset.user_with_deleted_scope` is useful in context and  technically correct, but maybe a bit less readable and "noun-like" than `asset.possibly_deleted_user`

sudara: I checked what paranoia and discard did, since they had to solve the same problems.

Maybe `asset.with_deleted_user` works? It sorta sounds like it's always a deleted user, but...

innatewonderer: I like your explanation. 100% agree.
`.with_deleted_user` will cause possible confusion considering we are using it for everyone :) 

`possibly_deleted_user` it is.

PR #752, opened 2019-09-13
Handful of frontend fixes

ofsound:
### Bugs

▹ shouldn't the compressed that didn't work! have a border still?

Screenshot 2019-09-13 12 55 31
Screenshot 2019-09-13 12 57 27

▹ don't show .user_card_content_divider if no location is set

Screenshot 2019-09-13 13 27 22

▹ admin user thumbnails are cropping strange & not scaling down

Screenshot 2019-09-13 13 08 53

▹ Edit/Playlist Long track names truncate awkwardly in the Available tracks right column, should break before the +

Screenshot 2019-09-13 13 11 17

Screenshot 2019-09-13 13 18 01

▹ Edit/Playlist Shorten the left column instructions, on thin 2 column views, the inconsistent line wrapping is bad, or at least link these divs together

Screenshot 2019-09-13 13 28 44

### Does anything special need to happen for deployment?

CSS changes should be superficial and only beneficial.

## "Ready For Review" checklist

These checklists are to help ensure the code review basics are covered. Consider removing to reduce noise.

* [ ] PR title accurately summarizes changes
* [ ] New tests were added for isolated methods or new endpoints
* [ ] I opened an issue for any logical followups
* [ ] If this fixes a bug, "Fixes #XXX" is either the very first or very last line of the description.

## Before code review *and after additional commits* during review.

* [ ] Update title and description to account for additional changes
* [ ] All tests green
* [ ] Booted up the branch locally, exercised any new code
* [ ] Percy changes are purposeful or explained
* [ ] Css changes are happy on mobile (via Percy is ok)

PR #754, opened 2019-09-16
Temporarily skip profile spam detection on update (Fix Master)

sudara:
Another valid user got caught in our spam filters on profile update — seeing that we have signups closed temporarily, I disabled the spam checking in https://github.com/sudara/alonetone/commit/d52cda05ae37a2bcd10e912113b73dfb428d87d8

Forgot to handle the specs though :)


sudara:
Percy handled in #752

PR #760, opened 2019-09-16
Add back to artist link where needed

ofsound:

* [ ] All tests green

### DIFF
--- a/app/assets/stylesheets/components/comments.scss
+++ b/app/assets/stylesheets/components/comments.scss
@@ -12,6 +12,13 @@
 /* RECENT COMMENTS PAGE */
 
 .comments_columns {
+
+  .columns_header {
+    max-width: 960px;
+    margin-left: auto;
+    margin-right: auto;
+  }
+  
   h2 {
     padding-left: 16px;
     color: $accent;
--- a/app/assets/stylesheets/components/comments.scss
+++ b/app/assets/stylesheets/components/comments.scss
@@ -14,11 +14,16 @@
 .comments_columns {
 
   .columns_header {
-    max-width: 960px;
+    max-width: 1032px;
+    box-sizing: border-box;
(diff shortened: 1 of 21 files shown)


ofsound:
I added the partial only on the /history page for now
It will also appear similarly on /tracks

In figma, it does appear on a playlist view, bottom left.

Screenshot 2019-09-16 15 18 25

there's a situation in mass edit where you have no navigation options to return anywhere, certain if this exact same back to artist link would be appropriate there, as sometimes you'd want to go back to the playlist you were editing track descriptions for.

sudara:
Just FYI the playlist view version is the original mockup, but we decided to consistently have it left for any page that needs it:

image

For create playlist, we'll have the cancel secondary button.
For edit playlist, we'll have the "view playlist" secondary button.

sudara:
I've plugged it in, but looks like we need css (and html?) love on 

* /sudara/comments
* /sudara/tracks
* /playlists

before we are mergeable.

1) I'm thinking we want to kill the hover over the artist name — it somehow makes the whole thing not feel like a button but more fiddley. 

2) Should the font be medium weight?

sudara:
Good call with the aligning — I checked in with smoofles, and this aligning seems to be unintentional: 

Messages Image(339800518)

ofsound:
@sudara 

I tried a medium font, i guess that matches a bit more closely the "view all >" style.

Removed the underline hover for now. View all has all text gray and then text color hovers to orange. 

I placed the button in those 3 other locations you mentioned.

ofsound:
@sudara I think if you conditionally add or not add this bit at the top of _asset_white, it will fix the layout issues that showed up with 2 buttons

<div class="track-content-header">
    <%= render partial: 'shared/back_to_artist' if @user.present? %>
</div>

PR #768, opened 2019-09-18
Improve "private" checkbox behavior by removing html ids

### LINKED ISSUE #578: Seekbar dissapears if you click other track while current one plays
Mainly problematic because when you click back on the track the seekbar stays hidden.

image

### LINKED ISSUE #747: Comment private toggle only works for 1st unique track on page
It looks like it fails if you try to toggle private on a "later instance" of the same track.

Nothing in web inspector.

git

sudara:
Fixes #747 
Fixes #578

## Context

 We were unable to make private comments on the 2nd or greater instance of assets on a page. 

In part this was due to reused ids due to asset caching — the asset was stuffed full of `asset.unique_id` which ironically is no longer unique when cached.

We rely on the caching, so the answer was to remove ids, as they do not provide a function.

In addition, I had to move around some of the click handling on the custom checkbox code to get the label + checkboxes to function again. I believe this is because the label no longer is auto-associated with the checkbox.

## QA


ofsound:
@sudara  – I cruised through these checkboxes, in comment assets and on the playlist edit. They seem to work fine on all platforms i tested, could click labels, correct values sent to the server..

ofsound:
@sudara – I think there might be a difference in how Firefox and Safari are handling the Private checkbox on the playlist edit page.

I think Firefox is updating the database when you click the input, whereas Safari you need to click  save changers.

I also think I've seen a situation in Firefox where I've clicked "save changes" and then hit Reload after that page has loaded, and seen the checkbox flip states, even when the only action was a page reload.

sudara:
Capybara still isn't clicking the private comment checkbox correctly...

PR #770, opened 2019-09-19
I761 Add user `show` action and view for admin namespace

innatewonderer:
#### About

It is very useful to see a user's page in an admin namespace. This page aggregates all relevant information for a user from an admin perspective (comments, assets, user info) to make sure an admin has complete information when deciding if a user is a spam or not.

This PR does NOT style the page, it mainly compiles the information (proof of concept).
I took existing `div` from `admin/user`, `admin/assets` and `admin/comments`.

- It has pagination set to 5 records (for comments and assets). Keep the page small
- It will display deleted user

#### Was anything tried that didn't work? Anything that reviewers should pay attention to or difficult or tricky that should be explained?

The main challenge turned out to be `routes`. At first, I added it as
```
resources :users, path: 'users/(:filter_by)', only: [:index, :show] do...
```
and didn't even notice any issues at first because it was hidden behind the existing functionality of `filter_by: email`. Instead of hitting `users#show` it would go to `users#index`. The routing here would treat `login` value not as `id:` parameter but as `filter_by`.

It surfaced when I added the specs. I fixed it by making an explicit route to `users#show` outside of `path: 'users/(:filter_by)'` scope.
Also, I had to give it a different name to differentiate `index GET` from `show GET`
```
admin_user GET      /admin/users/:id(.:format)                                                               admin/users#show
...
admin_users GET      /admin/users(/:filter_by)(.:format)                                                      admin/users#index
```
### FOUND BUG:
When adding specs to show comments that user gave to others, it appeared that it's not included in the `soft_delete_with_relations` scope.
So I added new `has_many: given_comments` association on a user to be able to access comments a user GAVE.
```
  has_many :given_comments, -> { order('comments.id DESC') },
    foreign_key: 'commenter_id',
    class_name: 'Comment',
    dependent: :destroy
```
#### Did I hear scope creep???
To bring the whole thing together I added few more things:
- Since we do not want to be able to do any action against assets whose user was deleted, we now hide the buttons if `data--user-deleted` is true
- Comments stumulujs controller haven't seen some love in a while. I refactored controller to follow the patterns established for other models
- One `gotcha`: since comments can have guest commenters `link_to_if` was refusing to cooperate
```
<%= link_to "Comment By: #{comment&.commenter&.login}", admin_possibly_deleted_user_path(comment.commenter.login), { filter_by: {commenter_id: comment.commenter_id } } %>
```
Even if I did a `link_to_if` per ruby docs gentle guidance, it would still give me an error 
```
ActionView::Template::Error (undefined method `login' for nil:NilClass):
```
So I fixed it by building an `if/else` block.

Added specs.

### Before code review *and after additional commits* during review.

* [ ] All tests green


innatewonderer:
@sudara Looks like Percy is failing on master updates? Would you take a look please?

app/commands/user_command.rb
```diff
@@ -39,14 +39,16 @@ def efficiently_delete_relations(time = Time.now)
     Playlist.joins(:assets).where(assets: { user_id: user.id })
             .update_all(['tracks_count = tracks_count - 1, playlists.updated_at = ?', time])
     Track.joins(:asset).where(assets: { user_id: user.id }).update_all(deleted_at: time)
-    Comment.joins("INNER JOIN assets ON commentable_type = 'Asset' AND commentable_id = assets.id")
```
innatewonderer: This seems to be getting user's comments on assets specifically, while `[comments].each do |user_relation|` would get both asset and blog comments...

How certain are we that it's a necessary extra step?

app/commands/user_command.rb
```diff
@@ -39,14 +39,16 @@ def efficiently_delete_relations(time = Time.now)
...
-           .joins('INNER JOIN users ON assets.user_id = users.id').where('users.id = ?', user.id).update_all(deleted_at: time)
-
     user.assets.update_all(deleted_at: time)
 
+    # comments given
+    user.given_comments.update_all(deleted_at: time)
```
innatewonderer: @sudara correct me if I'm wrong, but we might have wanted to leave any comments behind when a user is deleted. A quick check on my data showed a whole lot of comments without commenter. Plenty of them are not-spam. Plent-ier yet - spam.

I would like to suggest that we consider if we want to tighten the rules for what a guest user can do. Does it really matter to you as a musician to get an anonymous comment on your music?

Let's say we choose to leave comments behind when user is deleted, but let's NOT leave spam ones at least. 

Thoughts?

sudara:
@innatewonderer I think you might need to merge master in to get percy happy, a lot changed in the last days!

app/commands/user_command.rb
sudara: Comments originally were polymorphic in practice. These days, they aren't. I believe without cleaning the db, we might run into bugsnags, specifically on pages like `sudara/comments` unless we clean up the db/relationship.

app/commands/user_command.rb
sudara: > correct me if I'm wrong, but we might have wanted to leave any comments behind when a user is deleted

Historically we've kept true to our promise that we delete ALL the content you create when you delete your account. This has a secondary side effect of removing comments on other people's tracks where you might have said nice things, but I believe it's the cleanest solution currently.

> Does it really matter to you as a musician to get an anonymous comment on your music?

Yes, most family members or non-musician friends of musicians use guest comments quite regularly.

app/models/user.rb
```diff
@@ -89,9 +89,16 @@ class User < ApplicationRecord
   has_many   :playlists, -> { order('playlists.position') },
     dependent: :destroy
 
+  # this only covers comments received by the user
+  # not comments made by the user
   has_many   :comments, -> { order('comments.id DESC') },
     dependent: :destroy
 
+  has_many :given_comments, -> { order('comments.id DESC') },
```
sudara: I believe elsewhere in the code/specs, this is named `comments_made` — I think "made" is clearer than "given" but we should stick with one name. We could consider renaming the `comments` scope to `comments_received` to help the naming become more clear.

app/views/admin/users/show.html.erb
```diff
@@ -0,0 +1,138 @@
...
+    <div class="admin_column_right_inner box">
+      <h3 style="padding-left: 1em;">User</h3>
+      <div style="padding: 0 1em;">
+        <header class="admin_assets_header">
+          <div class="date_column">Date</div>
+          <div class="user_column">User</div>
+          <div class="track_column">Track</div>
+        </header>
+        <div class="admin_users_rows">
+          <div data-controller="admin--user" data-admin--user-deleted="<%= @user.soft_deleted? %>" data-admin--user-spam="<%= @user.is_spam? %>">
```
sudara: Divergence is almost guaranteed on these pages, as admin stuff is changing rapidly — we need to use layouts/partials for this work (vs. copy and paste).

app/controllers/admin/assets_controller.rb
```diff
@@ -15,17 +15,17 @@ def unspam
 
     def spam
       AssetCommand.new(@asset).spam_and_soft_delete_with_relations
-      redirect_to admin_assets_path(filter_by: :is_spam)
+      redirect_to admin_assets_path(filter_by: :is_spam) unless params[:no_redirect]
```
sudara: This `no_redirect` pattern here suggests to me that we're working around a problem and there's probably a clearer solution. Can you describe what the need is here? Are these controller methods not called via ajax in the new view?

sudara:
Excited for this!  Nice find on the comments scope. I think we could put that to use in comments_controller.rb

config/routes.rb
```diff
@@ -22,9 +24,6 @@
         put :unspam
         put :spam
       end
-      collection do
-        put :mark_group_as_spam
-      end
```
sudara: Nice call on removing this stuff!

app/views/admin/assets/index.html.erb
```diff
@@ -25,7 +25,9 @@
                 <%= user_image(asset.possibly_deleted_user, variant: :original) %>
               </div>
               <div>
-                <div><%= link_to asset.possibly_deleted_user.login, admin_users_path({ filter_by: asset.possibly_deleted_user.email }) %></div>
+                <div><%= link_to asset.possibly_deleted_user.login, admin_possibly_deleted_user_path(asset.possibly_deleted_user.login) %>
+                  <div class="deleted_at_date"><%== "Deleted #{local_time_ago asset.possibly_deleted_user.deleted_at}" if asset.possibly_deleted_user.deleted_at %></div>
+                </div>
                 <div><%= asset.possibly_deleted_user.ip %></div>
```
sudara: All this `possibly` stuff is making me look for a :possum: emoji hehe

app/controllers/admin/assets_controller.rb
innatewonderer: It was a really bad user experience when i'm on an `user#show` page and was being redirected to `assets#spam` if I wanted to spam some of user's assets.
I'm working on a user, I don't want to be redirected anywhere.

app/views/admin/users/show.html.erb
innatewonderer: Yes. I was thinking about that.

The only reason I left it as is was because I wasn't sure if we wanted some other layout for `admin/user#show` page or if a collection of three tables is acceptable.

When I was cleaning the UI there were elements (like links on user login in comments and assets table) that were useless and also tell us that it should have a different layout than assets#index. 

Thoughts? Maybe we could prioritize new user#show design soon-ish?

app/controllers/admin/assets_controller.rb
sudara: Gotcha! It seems like the spamming should be happening via ajax?...

innatewonderer: Well... we do some stimulusjs actions but then redirect to `assets/index(spammed)`[ in the controller](https://github.com/sudara/alonetone/blob/master/app/controllers/admin/assets_controller.rb#L16-L19)
```
    def spam
      AssetCommand.new(@asset).spam_and_soft_delete_with_relations
      redirect_to admin_assets_path(filter_by: :is_spam)
    end
```
Which made sense when we did it. I forsee us needed more bulk actions with minimum redirects, but for now, leave as is?

innatewonderer:
@sudara Will do. There were some changes done inside those partials that warranted they stay separately. But I think I can clean them up enough to be able to use same partial in both user#show and index pages.

sudara:
@innatewonderer Is this ready for re-review?

sudara:
Re-running build, as percy is showing missing fonts for "Home as Guest"

sudara:
Yay 🎈

PR #776, opened 2019-09-20
Don't migrate picture variants

Manfred:
They will be generated by Fastly and no longer by Active Storage.

Tested locally by migrating the seeds on the master branch and then switching to `mst/replace-paperclip-with-active-storage` where everything works.


sudara:
The percy glitch is old, going to merge.

PR #781, opened 2019-09-20
Update the playhead to use GSAP animations

sudara:
This PR moves waveform display to SVG (from canvas) generated on the backend (instead of the frontend) and animates it with GSAP, in lockstep with the playhead.

This means:

1) the animation is smooth (up to 60fps)
2) it can be paused on percy, meaning we can keep screenshotting playback scenarios without creating glitch like we are now
3) Initial waveform display should be quicker since it doesn't have to be populated by js

Current bug:

* After a single track ends, need to press play twice to get it to restart.

Followup refinements:

* Ensure waveform is cached


app/assets/stylesheets/components/single_and_playlists_player/playlist.scss
```diff
@@ -55,4 +55,7 @@
       margin-right: auto;
     }
   }
+  .player .waveform {
+    padding-top: 14px;
+  }
```
sudara: @ofsound I'm actually not sure why I had to do this, but it seemed necessary after switching to svg.

sudara:
@ofsound Mind kicking the QA tires on this one?

PR #782, opened 2019-09-23
New site header and user dropdown

### LINKED ISSUE #547: Find a place for search functionality on mobile
Maybe a separate page is needed, as the search bar can't fit into the header on mobile views.

### LINKED ISSUE #689: Implement new site header and user dropdown
1. Should look more clickable / like a menu when not interacting with it 
2. A way to get to user home that is more intuitive than clicking on name ("Your Stuff" item)

image

ofsound:
Fixes #689
Fixes #547

* [x] Replace logo with 200px one in design/
* [x] Replace admin button with edit profile
* [x] Take a crack at aligning position of profile image before/after animation
* [x] Dark version?
* [x] Mobile version


sudara:
I greedily went ahead and cleaned up some of the js here, adding the ability to hit ESC or click out of the dropdown to close it.

One option to accomplish that would be to have a stimulus event handler on our controller called `click@window->user-dropdown#close` — this would mean, any window click would call `close()` and then we could ask if the click happened in or out of the dropdown.

.....but I have problems with every single click in the app running through the user dropdown....

So I did a bit more manual solution of registering a one time click handler when `open()` fires. 

---

I also added a prototype little animation for the user image — it might be nice to add a bit of a bounce by seeing that image "grow" from 32 to 36px — however, the positioning would need some love. Maybe you could play a bit with it? If it doesn't work, it doesn't work...

sudara:
Also thinking we want a step darker gray for background of user image so it looks differentiated from the UI/chrome triangle.

sudara:
Event removing wasn't working, so I changed the event handling to be a bit more precise using this method:

https://medium.com/@photokandy/til-you-can-pass-an-object-instead-of-a-function-to-addeventlistener-7838a3c4ec62


## QA

* [ ] Open and click outside box, should close
* [ ] Open and click header of box, should close
* [ ] Open and click non-link in box (between edit/admin), should not close
* [ ] Open and click a link elsewhere in the UI, should close and visit new page
* [ ] Open and click back, should close and go back

sudara:
@ofsound IMO we should plow forward with this, work on dark/mobile and get it merged. Then we can ask samo for a couple tweaks / try 50px height.

ofsound:
@sudara 

I hit everything on your recent list. Check it out!

On mobile you can reveal the mobile search, but not close it due to space limitations?

Also a bit of flash when the mobile browser zooms to the focused state.

But I'm happy to have search on mobile for the first time ever!

sudara:
Looks like there's a legit percy difference, as the way to trigger the menu open has changed. Fixing...

image

sudara:
Not quite sure why these ended up being considered conflicts, but I'm going to go with the changes from this branch, let me know if i screw things up again :D 

image

sudara:
Mobile search is cool!!! 😎 Wondering if this approach is reasonable for the zoom thing https://stackoverflow.com/a/46254706/11699286

Issue #772, opened 2019-09-19
Was able to successfully choose an empty password via Password Reset, but then unable to login again with this empty password


ofsound:

sudara:
This is still actively a problem on master.

sudara:
This is the same problem and was a result of upgrading to Authlogic 5
https://stackoverflow.com/questions/2174082/force-validation-of-blank-passwords-in-authlogic

Basically, by default, validation of passwords don't occur updating the user model and they are empty, because we want to update other user's fields in a big long form, etc. So Authlogic sees blank passwords and ignores them.

However, when resetting passwords, we do not want to allow blank passwords, as the person is supposed to choose passwords. 

Therefore we actually need to tell Authlogic not to ignore the passwords.

See: https://github.com/binarylogic/authlogic/issues/693

PR #804, opened 2019-09-30
Add some styles to make errorExplanation at least halfway sane

### LINKED ISSUE #790: Not so fast young one page is borked
Try to change password to a new one that’s too few characters,
and you will receive error messages that aren't formatted!

ofsound:
Fixes #790

I imagine this might be temporary. I think these form requirements should be handled with Javascript next to the form itself. ( also, related issue is you can have a 0 length password and pass the validation #772 ) 

I don't know if this errorExplanation/notice paradigm has a future.

Before:

Screenshot 2019-09-30 13 24 09

After:

Screenshot 2019-09-30 13 31 33

### Does anything special need to happen for deployment?

I don't think so. Just adding some css rules to those specific divs.

* [ ] All tests green

### DIFF
--- a/app/assets/stylesheets/components/notifications.scss
+++ b/app/assets/stylesheets/components/notifications.scss
@@ -115,14 +115,29 @@ div.small_spinner {
   padding: 20px 30px;
   background-color: $background-color-one;
   width: 100%;
+  max-width: 960px;
+  padding: 0 $baseline * 1.5;
+  box-sizing: border-box;
   margin-right: auto;
   margin-left: auto;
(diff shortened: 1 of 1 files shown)


sudara:
Thanks for attacking — yeah, this current implementation is awkward because the old school flash message is separate from the explanation — and the validations are all on the backend.

There's not really a silver bullet for "translating" the backend validations (which we want to keep in place for data integrity purposes) to frontend. We could explore some of the solutions though: https://www.jorgemanrubia.com/2019/02/16/form-validations-with-html5-and-modern-rails/

PR #817, opened 2019-10-01
Update  /About, /Radio and /Users to a new page style that includes the top located .sub_nav structure

ofsound:
TODO

* [x] correctly handle mobile alignment of odd numbers of sub nav items
* [x] check out the top margin scheme again, maybe make a variable for that?
* [x] convert radio menu to have li.current 
* [x] add sub_nav html to the about page, add other about pages?
* [x] determine appropriate intermediate states between mobile and full width
* [x] re-write bottom parts of pages that used to have the sub nav in a left column?

* [ ] All tests green


ofsound:
I just added some fresh CSS to make the why section appear like the other sections.

Seems like there might need to be a distinction between another paragraph and what's a larger section like in FAQ.

I guess I wonder about adding more to this branch, as it's more about the subnav code?

sudara:
@ofsound Samo's advice here is to first "break" and do 2 rows of 3 for mid-sized devices like iPad portrait before going to 3 rows of 2.

sudara:
I guess the rules might also change if we only have 4 items to... Hmm.. We can add a class if that ends up being the case, but i think most of these will have 5-6 items.

Issue #834, opened 2019-10-07
Log Active Storage requests to a different location in development


sudara:
@manfred Any thoughts about this? Lots of noise between the signal, will be even more so if we flesh out seeds to be more usable for css work (engaging pagination).  

```
Started GET "/rails/active_storage/disk/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaDdDRG9JYTJWNVNTSnJkbUZ5YVdGdWRITXZNVEp6WW5JMmNXbHBaelZyWkdSNmNXWjVZMko2TldSNmFuQnJkUzlsT0dJMk9URmpNRFl5TWpKak0yTTBPR1V5WkRsbFpXWmlNMlF5WWpjM01HUTVOVE0yTXpCak9UZ3hOelF5TW1ZM05UZzRNV1ZsWWpRMU1qVmtNVGxqQmpvR1JWUTZFR1JwYzNCdmMybDBhVzl1U1NKUGFXNXNhVzVsT3lCbWFXeGxibUZ0WlQwaWFtVm1abVJ2WlhOemRXUmhjbUV1YW5Cbklqc2dabWxzWlc1aGJXVXFQVlZVUmkwNEp5ZHFaV1ptWkc5bGMzTjFaR0Z5WVM1cWNHY0dPd1pVT2hGamIyNTBaVzUwWDNSNWNHVkpJZzlwYldGblpTOXFjR1ZuQmpzR1ZBPT0iLCJleHAiOiIyMDE5LTEwLTA3VDE3OjQzOjM3LjQwMFoiLCJwdXIiOiJibG9iX2tleSJ9fQ==--6a3df61d6febae93ef342e4e621756ebb97c5b3b/jeffdoessudara.jpg?content_type=image%2Fjpeg&disposition=inline%3B+filename%3D%22jeffdoessudara.jpg%22%3B+filename%2A%3DUTF-8%27%27jeffdoessudara.jpg" for ::1 at 2019-10-07 19:38:38 +0200
Processing by ActiveStorage::DiskController#show as JPEG
[...]

PR #835, opened 2019-10-08
Silence logging for requests to Active Storage controllers in development

Manfred:
When Active Storage is using the `filesystem` storage variant and running in development mode we silence all requests to the Active Storage controllers because they make it hard to read the relevant output during development.

References #834.

### DIFF
--- a/lib/filtered_rack_logger.rb
+++ b/lib/filtered_rack_logger.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+# Works just like the regular Rails Rack request logger, but silences all requests to the
+# configured silenced request paths.
+#
+#   config.middleware.use(
+#     FilteredRackLogger,
+#     silenced: %w(/rails/active_storage)
+#   )
+class FilteredRackLogger < Rails::Rack::Logger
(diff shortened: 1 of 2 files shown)


lib/filtered_rack_logger.rb
```diff
@@ -0,0 +1,23 @@
...
+#   )
+class FilteredRackLogger < Rails::Rack::Logger
+  def initialize(app, silenced:)
+    super(app)
+    @silenced = silenced
+  end
+
+  def call(env)
+    if @silenced.any? { |path_prefix| env['PATH_INFO'].start_with?(path_prefix) }
+      Rails.logger.silence { @app.call(env) }
```
innatewonderer: Would it make sense to limit what kinds of requests get silenced? I'm worried about not being notified if something was actually wrong or broken.

Manfred: We're only silencing requests to `/rails/active_storage*` in development. These paths are Rails internals and only serve uploaded files. So if anything related images or audio is broken in development you can turn the regular logging middleware back on and see what's happening.

Issue #836, opened 2019-10-08
Move CI to Circle


sudara:
I've been happy, but lately we are seeing daily problems that are related to their infrastructure. I break CI enough without these being added into the mix, creating noise and delay.

Also it would be nice to run 5 concurrent builds ala circle/buildkite.

This is maybe not infrastructure, but apparently it's trying to run a branch that I just merged and then was deleted from github:
 
https://travis-ci.com/sudara/alonetone/builds/130981479

Manfred:
Sure, do you want me to help you with that somehow?

sudara:
Sure! This would be useful maybe after #793 

We keep seeing errors of many flavors, latest one here: https://travis-ci.com/sudara/alonetone/builds/131839631?utm_source=slack&utm_medium=notification 

I just don't know where it's most straightforward. Circle says we get to build 4 containers, not sure if that's by default or if I have to contact them. https://circleci.com/pricing/#faq-section-linux

sudara:
I actually am not very well educated on what the options are. I guess Circle would give us slightly more parallel jobs and I'm moving https://github.com/sudara/stitches to it, so we could do that.

sudara:
Wishlist:

1. Move to circle workflows 2.1 
2. Cache as much as possible / use orbs if it helps
3. Run feature tests in parallel to rest of suite

Manfred:
Do you still want to switch to CircleCI or should I close this issue?

sudara:
#1063 moved us to GitHub actions. Nice catch...

Issue #833, opened 2019-10-07
rake db:reset fails with `unknown attribute 'password'`


sudara:
This only happened the first time I used it, but thought I'd mention. Prob some obscure authlogic thing.

```
➜  alonetone git:(master) bundle exec rake db:reset
Dropped database 'alonetone_seeds'
Dropped database 'alonetone_test'
Created database 'alonetone_seeds'
Created database 'alonetone_test'
rake aborted!
ActiveModel::UnknownAttributeError: unknown attribute 'password' for User.
/Users/sudara/Sites/alonetone/db/seeds.rb:49:in `<top (required)>'
/Users/sudara/.rbenv/versions/2.6.1/bin/bundle:23:in `load'
/Users/sudara/.rbenv/versions/2.6.1/bin/bundle:23:in `<main>'
Tasks: TOP => db:reset => db:setup => db:seed
(See full trace by running task with --trace)
➜  alonetone git:(master) bundle exec rake db:reset
Dropped database 'alonetone_seeds'
Dropped database 'alonetone_test'
Created database 'alonetone_seeds'
Created database 'alonetone_test'
You can now sign in with: admin - testing123
You can now sign in with: moderator - testing123
You can now sign in with: musician - testing123
You can now sign in with: marieh - testing123
You can now sign in with: carole - testing123
You can now sign in with: petere - testing123
```

sudara:
Looks like this happens reliably on first time setups.

Manfred:
Authlogic checks if the database is loaded when we call `acts_as_authentic` and does nothing if there is no database. Authlogic is never initialized because `db:setup` runs `db:create` and `db:seed` in the same process. The ‘solution’ would be unload and reload the `User` class at the start of the seeds.

project-orcon:
> Authlogic checks if the database is loaded when we call `acts_as_authentic` and does nothing if there is no database. Authlogic is never initialized because `db:setup` runs `db:create` and `db:seed` in the same process. The ‘solution’ would be unload and reload the `User` class at the start of the seeds.

How would you do this - I've tried User.reset_column_information but I'm still getting the same error

sudara:
@project-orcon See the PR that closes this issue: https://github.com/sudara/alonetone/pull/920/files

PR #844, opened 2019-10-10
783-Admin users should search "display_name"

### LINKED ISSUE #783: Admin users search should search "display_name" too
https://alonetone.com/admin/users?filter_by=davidsen

should result in 
https://alonetone.com/kenneth
https://alonetone.com/ken

fadib:
Fixes #783 

### Iterate through the changes in this PR. Why did you implement them this way?

* Add a condition to also search column `display_name`
* Show `user.display_name` on the `admin#users#index#`

### DIFF
--- a/app/views/admin/users/index.html.erb
+++ b/app/views/admin/users/index.html.erb
@@ -34,7 +34,7 @@
                 <%= user_image(user, variant: :large) %>
               </div>
               <div>
-                <div><%= user.soft_deleted? ? user.login : link_to(user.login + " (#{user.display_name})", user_home_path(user.login)) %></div>
+                <div><%= user.soft_deleted? ? user.display_name : link_to(user.display_name, user_home_path(user.login)) %></div>
                 <div><%= user.email %></div>
                 <% unless user.soft_deleted? %>
                 <div><%= user.current_login_ip %>
--- a/app/views/admin/users/index.html.erb
+++ b/app/views/admin/users/index.html.erb
@@ -8,7 +8,7 @@
   <div class="admin_column_right">
    <div class="admin_search_field">
     <%= form_for 'filter_by', method: :get do |f| %>
-      <%= text_field_tag 'filter_by',nil, placeholder: 'Search by login or email' %>
+      <%= text_field_tag 'filter_by', nil, placeholder: 'Search by login or email' %>
     <% end %>
     </div>
     <div class="mini_paginator top_paginator">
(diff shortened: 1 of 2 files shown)


app/views/admin/users/index.html.erb
```diff
@@ -34,7 +34,7 @@
                 <%= user_image(user, variant: :large) %>
               </div>
               <div>
-                <div><%= user.soft_deleted? ? user.login : link_to(user.login, user_home_path(user.login)) %></div>
+                <div><%= user.soft_deleted? ? user.login : link_to(user.login + " (#{user.display_name})", user_home_path(user.login)) %></div>
```
sudara: Our admin is not the prettiest situation in the moment, but I don't think there's enough horizontal space to add display_name here. We could just use `user.display_name` instead of `login` — this is the pattern used across the site: https://github.com/sudara/alonetone/blob/master/app/models/user.rb#L80

image

sudara:
Kind of strange Travis didn't run on this PR, wonder if it's just a glitch or something to do with submitting from a fork.

sudara:
It looks like "build pushed pull requests" was off, probably because it was doing duplicate work with building branches. Not quite sure how best to resolve, but I've turned it back on from now. If you end up wanting to contribute a few more PRs, we can eventually give you commit access.

sudara:
https://docs.travis-ci.com/user/pull-requests/#double-builds-on-pull-requests IMO, another reason for #836

fadib:
> Welcome! And thanks for this!
> 
> Did you get all setup locally ok? Feel free to add issues if you run into anything not ideal or not clear, especially with setup, as we recently invested in seeds, etc.

I stumbled error per this issue #833 but now all is good

app/views/admin/users/index.html.erb
```diff
@@ -34,7 +34,7 @@
                 <%= user_image(user, variant: :large) %>
               </div>
               <div>
-                <div><%= user.soft_deleted? ? user.login : link_to(user.login, user_home_path(user.login)) %></div>
+                <div><%= user.soft_deleted? ? user.display_name : link_to(user.display_name, user_home_path(user.login)) %></div>
```
sudara: Sorry, I should have been more clear here — the `link_to` still needs to use `login`, that's the unique key that we find by.

sudara: Whups ignore that comment, I misread!

PR #851, opened 2019-10-13
Improve Edit/Create Playlist UX 

### LINKED ISSUE #690: Create Playlist UX
## Design Needs

* Info box 
* Clear private / public toggle
* Button states

## Workflow
1. "create album or playlist" goes directly to playlist/edit
  * Info box describing what a playlist is
  * No ability to add tracks yet
  * Button says "create playlist" 

2. Playlist/edit (for the first time after create)
  * No info box
  * Ability to add tracks shows up
  * Option to upload track(s) / zipfile
  * Minimize the metadata / title stuff
  * Clearly marked as private
 
3. Returning to Playlist/edit 
  * Everything is seen, including upload

Private / View / Save button states: 
  * Disabled Button says "Save" until something changes, then it enables
  * "View" button always present
  * Very clear private/unpublished state

Bandcamp example: 

Screenshot 2019-08-07 12 02 48

sudara:
Closes #690 

## Changes made

* [x] Playlists create now uses new form
* [x] Ability to toggle private is hidden on create
* [x] Ability to independently upload image goes away in favor of using main "save" button
* [x] Added "back to your profile" on create and "back to your playlist" on edit
* [x] Implement new "private" banner as shown in figma
* [x] Change text of private banner to reflect hidden vs. private
* [x] Don't display private checkbox if less than 2 tracks
* [x] Get rid of file selector? Or at least make the image clickable to upload. 

## CSSTUNE

* [x] Style "Choose Artwork..." ala figma
* [x] Pick a style for "Remove Artwork" for now... guess we should have a bit of space between?
* [x] Implement button styles as seen in figma https://www.figma.com/file/YdjrVsNumbBsWVo82Wje2h6N/alonetone-white-theme?node-id=950%3A4218
* [x] Implement "Upload Artwork..." and "Remove Artwork" styles
* [ ] Implement "figma style" border dashing around the playlist cover.
* [ ] In playlist cover view, need private indicator when track has 0 tracks
* [ ] Check mobile, responsive? Decent ordering?

## OTHER 

* [ ] Blank slate from Alec

## Open questions

* [ ] Should we not allow people to create a playlist until they've uploaded tracks?
* [ ] Should we "collapse" the main form after create to clearly direct people to add tracks?


sudara:
image

sudara:
@ofsound Could csstune and qatune have a gander at the current version? Looks like there are still some issues:

image

But also, I could use an objective eye to understand if there's still some UX confusion here.

1. I made it so that if you remove tracks to less than 2, things look like they will go private again. However, on the backend that's not currently true. Do we care about enforcing that?

2. If you start with an empty "hidden" playlist, then add 2 tracks, the default is it's still private. I'm not sure it's totally clear that you need to uncheck "private" and then hit save. Any ideas?

I feel like this is workable, but almost feels like we are missing a "public" banner...

sudara:
I added some code to finally get rid of the file field and replace with the "Upload Artwork" link that inserts a preview when you choose a file.

That is all functional, however I also added a "Remove Artwork" link which still needs some #jstune to be happy.

ofsound:
@sudara 

Maybe we could re-use the Padlock overlay when viewing a private playlist. This banner is sort of a eyeful, but I guess for showing up on all views of a playlist, maybe showing up here under the title in the sidebar?

Screenshot 2019-10-18 15 53 32

Screenshot 2019-10-18 15 52 28

ofsound:
_I made it so that if you remove tracks to less than 2, things look like they will go private again. However, on the backend that's not currently true. Do we care about enforcing that?_

I guess that was so some Benjamins-esque character wouldn't release a bunch of singles with album art and take over the homepage? Maybe that restriction is outdated? not sure! I'd guess the reason to dynamically push that change into the backend if it stays as it is now, would be to make the "This playlist will be hidden until it has at least 2 tracks" statement true?

_If you start with an empty "hidden" playlist, then add 2 tracks, the default is it's still private. I'm not sure it's totally clear that you need to uncheck "private" and then hit save. Any ideas?_

I agree this might be unexpected behavior. Maybe the banner could say ~ Your playlist is now public and available to anyone at this link : http://alonetone.com/user/playlist 

_We probably shouldn't let a user create a playlist until they have tracks right now. Not sure how to do this except by removing the ability to create them from the user dropdown? We could redirect them on clicking it to go to the upload form..._

I think that's a better solution than not showing them the button. Seems like sometime in the future we may have the flow of creating a playlist first and then uploading tracks into it.

sudara:
@ofsound I might merge this even though it isn't "done" so that I can pull the trigger on legacy. 

I put the dashed border back around the images, which I believe was intended (not just svg).

Your orange mockup for "private" feels appropriate for the larger playlists — it should be very obvious and not a text-less visual cue. We could also do something like we had for the "warning: this cover image is lofi", which was a bar at the bottom of the cover. But I'll open an issue for that as a followup so we can close this out.

PR #854, opened 2019-10-17
Add extra form validation on password and password_confirmation.

innatewonderer:

* [x] All tests green

### DIFF
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -83,8 +83,8 @@
   <div class="static_content">
   <div id="edit_password" class="static_content" style="display:none">
 		Enter your desired new password twice, please:<br/>
-		<%= f.password_field :password %><br/>
-		<%= f.password_field :password_confirmation %><br/>
+		<%= f.password_field :password, required: true %><br/>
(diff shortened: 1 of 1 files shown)


sudara:
@innatewonderer Would you mind adding some description to this PR? Does this resolve #772? I'm guessing `require_password` is an authlogic thing?

Does password validation work here generally now? It looks like the minimum character limit is 8, according to this code, would be nice to confirm that all works as expected: https://github.com/sudara/alonetone/blob/master/app/models/user.rb#L46-L57

innatewonderer:
Apologies, @sudara. The description never migrated from the commit message.
The validation does work. But it looks when password string comes in as nil or empty, it doesn't get caught in the length but in `presence: true`.  And if we add `presence: true` to validation, we won't be able to do partial updates (which is why I think it was omitted from the start).

sudara:
@innatewonderer OK, just to be clear: you tested both both blank passwords and other passwords < 8 characters on the form and now that `require_password` is on the form, they are properly rejected?

sudara:
I think I misunderstood or acted too fast here. We should have had specs, this is still a problem on master. See #772

Issue #627, opened 2019-06-24
On user/tracks/ I think Most Commented/Fav'd/Listened Tracks aren't accurate


ofsound:
On /sudara  Suffocate tops these lists

sudara:
The core problem here seems to be these queries seem to only be looking at a subset of the assets.

https://github.com/sudara/alonetone/blob/master/app/controllers/assets_controller.rb#L29-L32 

Perhaps only the last 30 days? Or maybe the order isn't actually correct since `@user.assets` already has an order?

<img width="527" alt="image" src="https://user-images.githubusercontent.com/472/67504394-93139f80-f689-11e9-928b-b26c62053fc3.png">

To reproduce in tests, you could create some assets with created_at over a year old, with high favorites count, high listens count, etc — and verify those top the list.

Also we should make the lists 5 tracks in length, 10 seems too long!

innatewonderer:
Looks like `.order()` is broken:
```
2.6.3 :006 > @user.assets.most_commented.limit(10).map(&:comments_count)
  Asset Load (2.7ms)  SELECT `assets`.* FROM `assets` WHERE `assets`.`deleted_at` IS NULL AND `assets`.`user_id` = 1 AND (comments_count > 0) ORDER BY assets.id DESC, comments_count DESC LIMIT 10
 => [3, 5, 2, 2, 1, 1, 1, 1, 1, 1]
2.6.3 :007 > @user.assets.most_listened.limit(10).map(&:comments_count)
  Asset Load (1.8ms)  SELECT `assets`.* FROM `assets` WHERE `assets`.`deleted_at` IS NULL AND `assets`.`user_id` = 1 AND (listens_count > 0) ORDER BY assets.id DESC, listens_count DESC LIMIT 10
 => [3, 5, 2, 2, 1, 0, 1, 0, 0, 0]
```
Both of these methods are scopes and contain `.order('### DESC') `.

It also affects `hottest`, which at first I was wondering if `order` is not ordering floats correctly, but then found that it seems to fail to order any of them. 

I found at least one reported issue in Rails gem for version 6.0 
https://github.com/rails/rails/issues/37376
P.S. I did confirm that `limit(10)` does not impact the result.

innatewonderer:
[...]

PR #873, opened 2019-10-26
I627 fix user/tracks order in tracks lists

innatewonderer:
### [Bug] #627 

If we chain `order` method in Rails, it will order by first value and then order by second value what's available. So since for associations first order is by`id` and it's always unique, it does not (EVER) actually order by the second value :/ So none of user's assets get ordered in their lists.

I decided to add at least some specs (on a `model` since we can't really test for order of assets as part of a controller spec) to have at least SOME check on those lists.

* [x] All tests green

### DIFF
--- a/app/assets/stylesheets/application/tracks.scss
+++ b/app/assets/stylesheets/application/tracks.scss
@@ -207,12 +207,6 @@
       -webkit-box-sizing: border-box;
     }
   }
-
-  &.descriptionless {
-    .seekbar {
-      background-color: $tab_box_silver;
-    }
-  }
 }
 
 div.track_options {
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/application/tracks.scss
```diff
@@ -207,12 +207,6 @@
       -webkit-box-sizing: border-box;
     }
   }
-
-  &.descriptionless {
-    .seekbar {
-      background-color: $tab_box_silver;
-    }
-  }
```
sudara: I think this whole file is old dark theme stuff (no longer accessible).

Issue #829, opened 2019-10-06
Remove Legacy Dark Theme


sudara:
* [ ] Add message to users with `use_old_theme` flag
* [x] Remove theme switcher in footer
* [ ] Optional: Migrate users to beta dark theme

After 

* [ ] Remove all old css / js 
* [ ] Rename layouts
* [ ] Clean up old image variants and tests in playlists_helper 
* [ ] Rename partials

PR #877, opened 2019-10-27
Remove legacy theme

### LINKED ISSUE #393: Kill Greenfield
* [x] Weep tears
* [x] Screenshot for posterity
* [x] Add git tag
* [x] Figure out what to do about PlaylistDownload (keep and chat w/ @Manfred about it or toss for now)
* [ ] Migrate content from `Greenfield::Post` to the relevant tracks
* [ ] Remove route constraint
* [ ] Remove assets
* [ ] Remove gem

sudara:
Fixes #393 and #829  

Changes in this PR are dauntingly huge, but they boil down to a handful of things:

0. Greenfiled engine is removed
1. Old theme views are removed and all `_white.html.erb` views are renamed to follow rails conventions.
2. All legacy `ImageVariant`s are removed in specs
3. All legacy theme  js/css is removed
4. All `white_theme_enabled?` calls are removed
5. All `theme_name` additions to cache keys are removed
6. There was lots of messy logic around the `assets/asset` partial that was no longer needed, so I made things a bit more flexible/sane by introducing the `normal_playback` layout for assets (sometimes assets are in other contexts, such as in a playlist sortable, etc).

This is a brutal but not complete removal of legacy/greenfield. The following are open:

* Renaming `User#white_theme_enabled` to be useful for the upcoming new dark theme toggle.
* Removal of old forum related models 
* Removal of greenfield models (waiting on a manual small migration by yours truly)
* It's possible I missed some image related logic that is no longer necessary
* There's bound to be a few remaining views that I didn't manage to run into
* All the old images are still in place, I will probably move them into an archive location in the git repo


app/controllers/assets_controller.rb
```diff
@@ -9,7 +9,7 @@ class AssetsController < ApplicationController
   before_action :require_login, except: %i[index show latest radio listen_feed]
   before_action :check_new_user_abuse, only: %i[new create]
 
-  etag { "#{white_theme_enabled?}/#{current_user&.id}" }
+  etag { "#{current_user&.id}" }
```
Manfred: Doesn't even have to be a string at this point.

app/controllers/assets_controller.rb
```diff
@@ -170,7 +162,7 @@ def update
         redirect_to user_track_url(@asset.user.login, @asset.permalink)
       else
         flash[:error] = "There was an issue with updating that track"
-        render action: white_theme_enabled? ? "edit_white" : "edit"
+        render action: "edit"
```
Manfred: Or `render :new` as the cool kids do nowadays.

app/models/image_variant.rb
```diff
@@ -5,17 +5,12 @@ class ImageVariant
   VARIANTS = {
     small_avatar: 80,
     medium_avatar: 96,
+    card_avatar: 440,
     large_avatar: 960,
-    playlist_avatar: 440,
     playlist_card: 440,
     playlist_cover: 1248,
```
Manfred: I would keep the number of sizes to an absolute minimum, in the end it's going to be faster for users to download a slightly bigger variant once than having to download 4 size variants of the same thing. Especially if they end up downloading the large variant at some point anyway.

sudara: These are pretty much the current minimum, though we could probably remove the 80px one. The large_avatar and playlist_cover only show on one page and warrant an additional download.

app/controllers/assets_controller.rb
sudara: Good call.

PR #878, opened 2019-10-28
I612 order published playlists

### LINKED ISSUE #612: Sort playlists by `published_at` on home
Double check and ensure it's not possible to publish and set the date more than once.

innatewonderer:
#### Bug: `published_at` isn't set on update
FIxes #612

- When update action is triggered, `published_at = Time.now` would not actually persist to save when `update` is complete. `self.` took care of that.
- I checked all assignments in `playlist.rb` and fixed the same issue on `is_mix`
- Per Sudara's comment below, I added a spec to ensure playlist cannot be published if there are less than 2 tracks. You COULD publish, so I fixed that with `can_be_public?` method. Adding follow up issue #882 to cleanup `before_update` callback. There is way too much magic going on inside that callback, some of it touches the same domain and could cause bugs that are tough to troubleshoot.

### Was anything tried that didn't work? Anything that reviewers should pay attention to or difficult or tricky that should be explained?
At first, I thought the issue was somewhere in failed callbacks. But it didn't appear to be the case.

* [X] PR title accurately summarizes changes
* [X] New tests were added for isolated methods or new endpoints

* [x] All tests green


innatewonderer:
@sudara Percy is very unhappy on this one... doesn't seem related to my work.

sudara:
This looks good — I'm wondering what to do about all the existing data. We probably need a migration that sets any `published_at` timestamps to the last `updated_at`.

Also, we might want to consider an index here on `published_at`, depending on the query and what it's currently using. Existing indexes:

```
    t.index ["permalink"], name: "index_playlists_on_permalink"
    t.index ["position"], name: "index_playlists_on_position"
    t.index ["user_id"], name: "index_playlists_on_user_id"
```

Looks like maybe the position/user_id could be a compound index, too, depending on what the playlists query on the user home page looks like.

spec/controllers/playlists_controller_spec.rb
```diff
@@ -146,6 +146,53 @@
...
+      expect(playlists(:jamie_kiesl_loves).reload.published_at).not_to be_nil
+    end
+
+    it "should set is_mix to true if is_favorite?" do
+      put :update, params: {
+        id: playlists(:jamie_kiesl_loves).id,
+        permalink: 'jamie-loves',
+        user_id: user.login,
+        playlist: {
+          is_favorite: true
```
sudara: Bit confused about this — we shouldn't actually be able to do this, so you might have found a bug. There are only 1 favorites playlist per user, it's sort of a "magic" case of playlists.

spec/controllers/playlists_controller_spec.rb
```diff
@@ -146,6 +146,53 @@
...
+      expect(playlists(:jamie_kiesl_loves).private).to be(true)
+
+      put :update, params: {
+        id: playlists(:jamie_kiesl_loves).id,
+        permalink: 'jamie-loves',
+        user_id: user.login,
+        playlist: {
+          private: false
+        }}
+      expect(playlists(:jamie_kiesl_loves).reload.published_at).not_to be_nil
```
sudara: Setting `published_at` should only work if the playlist has 2 tracks. 

Can you add a test that < 2 tracks does not set published_at (it's not an option in the UI, for example).

sudara:
> @sudara Percy is very unhappy on this one... doesn't seem related to my work.

We ran out of percy screenshots, I had to start paying to get things rolling again. I'll try and refresh master so the comparison is right. Plus, I still have that last piece of glitch....

sudara:
Ahh, it looks like you might not have pulled latest master before starting the branch, hence the old comparison. I just pushed a merge.

spec/controllers/playlists_controller_spec.rb
sudara: Speaking of "magic" — the playlist creation for favorites happens in a very obscure location: https://github.com/sudara/alonetone/blob/master/app/models/track.rb#L40

We could create a followup where we make that more explicit and maybe even have a favorites_controller that behaves a bit more RESTful. That code is so old...

spec/controllers/playlists_controller_spec.rb
innatewonderer: Good idea. Will add a spec.

innatewonderer:
> I'm wondering what to do about all the existing data.
I was thinking the same thing. I couldn't think of any better way than `a migration that sets any published_at timestamps to the last updated_at`.

spec/controllers/playlists_controller_spec.rb
innatewonderer: OK. Now I'm confused :)
https://github.com/sudara/alonetone/pull/878/files#diff-8b4a98db7ad82717ddadf5da294a2c1aR139
What does this do then ^?

sudara: Great Q! In the UI we used to have a stronger definition of mixes (tracks from more than one artists) and albums (tracks from one artist only). 

Favorites, being inherently a mix, were marked as mixes. 

These days we decided not to implement mixes yet. But we may add them soon if artists ask for it again. So although this code was old and broken, it’s nice to keep up the distinction on the backend. 

Since it hasn’t been functional, maybe we need to backfill mixes too, just so the data is clean.

app/models/playlist.rb
```diff
@@ -128,15 +129,21 @@ def self.latest(limit = 5)
   end
 
   def ensure_private_if_less_than_two_tracks
-    self.private = true if !is_favorite? && (tracks_count < 2)
+    self.private = true if !is_favorite? && !can_be_public?
```
sudara: Do you know why this wasn't working before? It doesn't look like the logic actually changed (though I like it being broken out!)

spec/controllers/playlists_controller_spec.rb
sudara: This still needs fixing — `is_favorite: true` isn't something that we can pass in the playlist params...

spec/controllers/playlists_controller_spec.rb
```diff
@@ -146,6 +146,66 @@
...
+
+    it "should set is_mix to true if there is a mix of tracks" do
+      expect(playlists(:jamie_kiesl_loves).tracks.count).to eq(4)
+      expect(playlists(:jamie_kiesl_loves).is_mix).to eq(nil)
+      put :update, params: {
+        id: playlists(:jamie_kiesl_loves).id,
+        permalink: 'jamie-loves',
+        user_id: user.login,
+        playlist: {
+          is_favorite: false
```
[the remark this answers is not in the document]
innatewonderer: Hmmm, I see what you are saying, but `add_track` doesn't seem to be actually triggering an `update` action. Which leads me to believe that we might be making some unnecessary calls (in that method) during each actual update action.

Is it possible you are thinking of the good old days when add_track wasn't async? I was playing around `playlist/edit` page and you don't actually have to click `Save` to add a track with `+`. 

What do you think about making a ticket to clean up `set_mix_or_album` (make it an explicit action)? I could add it here, but it'd blow this PR up a bit.

app/models/playlist.rb
innatewonderer: It was working, just needed this same check to ensure we are not allowing publishing if there is less than two tracks.
It might be overkill (considering that you said that this feature is disabled at a view level and we are not allowed to publish a playlist if there is less than 2 tracks), but it doesn't hurt to have a service level check.

spec/controllers/playlists_controller_spec.rb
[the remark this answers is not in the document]
sudara: Happy to clean up mixes here if it makes sense, otherwise feel free to back out of those mix specs and changes. Track add/removing was always async. Not sure what happened here, let's should just make it nice...

One option is to use `before_add` https://guides.rubyonrails.org/v5.1/association_basics.html#association-callbacks — model callbacks might feel too magical, so we can also call `@playlist.set_mix_or_album` explicitly from the controller, since tracks are added and removed by 2 actions only.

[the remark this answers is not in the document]
innatewonderer: Few things I've tried that didn't work when trying to write a spec for this:
- When adding a first favorites track, new playlist is created, but it doesn't trigger `before_update` action
```
  Playlist Create (14.4ms)  INSERT INTO `playlists` (`title`, `user_id`, `created_at`, `updated_at`, `permalink`, `is_favorite`) VALUES ('innatewonderer\'s favorite tracks', 34856, '2019-11-05 05:16:26', '2019-11-05 05:16:26', 'innatewonderers-favorite-tracks', TRUE)
```

- when I favor a track and it makes it to my favorites playlist, but it triggers `"/toggle_favorite?asset_id=111783"`
```
Started GET "/toggle_favorite?asset_id=111783" for ::1 at 2019-11-05 00:05:21 -0500
Processing by UsersController#toggle_favorite as JS
```
The interesting thing here is that it DOES trigger `Playlist Update All (0.5ms)...`.
```
  Playlist Update All (0.5ms)  UPDATE `playlists` SET `playlists`.`tracks_count` = COALESCE(`playlists`.`tracks_count`, 0) + 1, `playlists`.`updated_at` = '2019-11-05 05:21:57' WHERE `playlists`.`id` = 20219
```
However, it skips all the callbacks. My current version is that counter_cache functionality skips callbacks. I didn't find anything in [the docs](https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-counter-cache). Found this in the [rails codebase](https://github.com/rails/rails/blob/a592e87a278cec10569d854dd5ece22179471bc0/activerecord/lib/active_record/counter_cache.rb#L65):
> It simply does a direct SQL update for the record
> with the given ID, altering the given hash of counters by the amount
> given by the corresponding value

- In order to be able to trigger `add_track`, I click the `+` on playlists/edit but the only options there are user’s own tracks (so we can’t really add other tracks that are is_favorite

Proposal for `is_mix` flag:
- Move `self.is_mix = true if is_favorite?` directly to when a new favorite's playlist is created. Then my prooving spec can cover creation of new favorites playlist and confirm that the flag is explicitely set.
- Remove `self.is_mix = true if tracks.present? && tracks.count > tracks.where('user_id != ?', user&.id).count` completely. It's a check that's performed on every `playlist.update` and not necessary
- Remove `set_mix_or_album` callaback

Thoughts?

app/models/playlist.rb
sudara: Ah cool, didn't realize it was in use in more than one place.

spec/controllers/playlists_controller_spec.rb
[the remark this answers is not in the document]
sudara: > the only options there are user’s own tracks (so we can’t really add other tracks that are is_favorite

The new playlist edit UI doesn't yet expose the other options, but we will be rolling those out (and have it designed) so we should keep the mix/album distinction on the backend.

> It's a check that's performed on every `playlist.update` and not necessary

You are right that a favorites playlist is always a mix. However, on normal track add, we do need to check if a playlist contains tracks from a user other than the user making the playlist and set `is_mix` to be true. +1 to removing the callback though! We can do that in `add_track`/`remove_track` (which ideally get moved into a new tracks_controller as `create`/`destroy` as those actions haven't been updated since REST was a thing!!).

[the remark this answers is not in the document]
innatewonderer: > add_track/remove_track (which ideally get moved into a new tracks_controller as create/destroy 
Another ticket #894

spec/fixtures/tracks.yml
```diff
@@ -95,6 +95,7 @@ jamie_kiesl_loves_1:
   position: 1
   created_at: <%= 1.day.ago %>
 
+# don't add any more tracks to this playlist
```
sudara: It would be helpful to have a "why" here...

db/migrate/20191029022906_populate_is_mix_on_playlist.rb
```diff
@@ -0,0 +1,11 @@
+class PopulateIsMixOnPlaylist < ActiveRecord::Migration[6.0]
+  def change
+    counter = 0
+    Playlist.all.map do |playlist|
+      playlist.set_mix_or_album
```
sudara: I think you removed / renamed this method, does it need to change here?

sudara:
@innatewonderer can you clarify what's happening with mixes now? It looks like 1) there's a migration to identify old mixes and 2) favorites are marked as a mix. 

Will we need to add new logic when we uncomment the ability in the UI to add tracks from other users to a playlist?

app/models/playlist.rb
```diff
@@ -125,27 +126,34 @@ def self.latest(limit = 5)
...
   end
 
   def check_for_new_permalink
     generate_permalink! if title_changed?
   end
+
+  def consider_a_mix?
+    return true if is_favorite?
+
+    tracks.present? && tracks.count > tracks.where('user_id != ?', user&.id).count
```
sudara: Something like `tracks.pluck(:user_id).uniq > 1` might be an upgrade here, more direct and less queries.

app/models/playlist.rb
```diff
@@ -5,12 +5,13 @@ class Playlist < ActiveRecord::Base
 
   scope :albums,           -> { where(is_mix: false).where(is_favorite: false) }
   scope :favorites,        -> { where(is_favorite: true) }
-  scope :for_home,         -> { select('distinct playlists.user_id, playlists.*').recent.only_public.with_preloads }
+  scope :for_home,         -> { select('distinct playlists.user_id, playlists.*').recently_published.only_public.with_preloads }
   scope :include_private,  -> { where(is_favorite: false) }
   scope :mixes,            -> { where(is_mix: true) }
   scope :only_public,      -> { where(private: false).where(is_favorite: false).where("tracks_count > 1") }
```
sudara: Seems like`where("tracks_count > 1")` is no longer needed here if we are guaranteeing a playlist can't be published unless it has 2 tracks.

app/models/playlist.rb
```diff
@@ -5,12 +5,13 @@ class Playlist < ActiveRecord::Base
 
   scope :albums,           -> { where(is_mix: false).where(is_favorite: false) }
   scope :favorites,        -> { where(is_favorite: true) }
-  scope :for_home,         -> { select('distinct playlists.user_id, playlists.*').recent.only_public.with_preloads }
+  scope :for_home,         -> { select('distinct playlists.user_id, playlists.*').recently_published.only_public.with_preloads }
```
sudara: I forgot about this too, but we probably need to change `playlists_controller#all` to have the same logic (sorting by published vs. recent). I believe then we can kill the `recent` scope below.

db/migrate/20191029022906_populate_is_mix_on_playlist.rb
innatewonderer: good catch!

app/models/playlist.rb
innatewonderer: Oh this is such a sweet trick. I love it.

innatewonderer:
> @innatewonderer can you clarify what's happening with mixes now? It looks like 1) there's a > migration to identify old mixes and 2) favorites are marked as a mix.

So we currently have a method that has a complete check to see if a playlist is a mix, but it is ONLY run during creation of favorites playlist. Once we restore Mix UI functionality, we can incorporate the method into it.

> Will we need to add new logic when we uncomment the ability in the UI to add tracks from other users to a playlist?

Yes. It was a callback on an update, but we found it's inefficient. Here are two options I'm thinking of now:
- add it as part of `add_track` refactoring in #894. My reservation here is that maybe we don't want it to be done on EVERY `add_track` request.
- If UI is more specific (for example, allowing a user to choose which playlist to add a track to and it's telling that this playlist is a mix (just like with favorites), then we could add the logic there (then it only runs once during creation of a Mixed album. This option might be overthinking, but I think once we know the UI of this feature better, it will be obvious where a check should happen.

Thoughts?

sudara:
> So we currently have a method that has a complete check to see if a playlist is a mix

It looks like the code is right, but just to clarify for the future, a playlist is always a mix by definition: it's a collection of tracks from favoriting tracks around the site. We don't need to check at creation or any other time, we just need `is_mix` to be true. Which is how the code looks, so I'm going to merge yay!!!

> Once we restore Mix UI functionality, we can incorporate the method into it.

Ok, that's not my preference, but I'm ok with deferring adding back this feature if it's helpful in some way. It should definitely be added back during add_track/remove_track work in #894. We want the backend to be doing the expected thing, regardless of the fact that we've temporarily hidden the ability on the frontend.

sudara:
🥇

PR #893, opened 2019-11-06
New edit profile page

### LINKED ISSUE #695: Implement Edit Profile
image

sudara:
Closes #695 

Open Questions:

* [ ] Should we ajax submit all these forms so that the flash messages occur in place? I'm leaning towards that *except* it means we need to a) have a spinner? b) feed useful errors into the stimulus flash controller. Are there forms where > 1 error might need to display at once?

Sudara

* [x] Get Rails forms roughed in
* [ ] Stimulus ajax message wording needs to be improved
* [ ] trim http/https from website field
* [x] migration to remove http/https from existing website attributes
* [ ] add https:// to the input field for website?
* [x] Stimulus Toggles that adjust individual settings and flash success
* [ ] Stimulus controller so browse button, default profile image and existing profile images opens file selector
* [x] Accept nested attributes for `@profile` in user forms
* [ ] should the bounding box be 940px or 960px?
* [x] Better description, labels? for Change Password section

Ben
* [ ] <hrs> in settings don't line up
* [ ] Change login / change password submit buttons need love
* [ ] Clean up error messages
* [ ] Change your login mobile collapse
* [ ] iTunes link line breaks in FF
* [ ] Long ajax messages break out of box
* [x] Toggles for settings
* [x] DT Form fields ala figma
* [x] Country select needs to be styled like a text input
* [x] Check size/color on "Profile Picture" text
* [x] "Save Profile" needs to be an input:submit via `f.submit` not an `<a>` — I added input[type=submit] to `edit_profile_top_left` but the formatting didn't completely change (gray button, no rollovers)
* [x] "Update Links" button should be same style as "Save Profile" and clear:both
* [ ] When a field fails validation (like password confirmation) the div is wrapped by rails, we need to highlight the field ala figma
* [x] "Danger" and "Warning" text on DT still has white background
* [x] Cancel button should be gray
* [x] Cancel and "Save Bio" button stack on top of each other in mobile view

image


sudara:
@ofsound Maybe you could have another round looking at those submit buttons, toggles, DT forms oh my. 

My next todos are 1) figure out if we can use stimulus ajax flashes 2) if not, how will we show validation errors in the middle of the page, etc.

sudara:
@ofsound I tried to merge master in, but I think there's some conflict with form.scss, so I'll leave it to you!

sudara:
@ofsound The conflict looked easy to resolve, so i grabbed it. I updated the description with the todos left on this branch. I'd love to close this one out!

sudara:
@ofsound I'm merging this today. Looks like we have one other css thing, which is the hrs not lining up on settings:

image

sudara:
Looks like we have some error handling to clean up too:

image

PR #895, opened 2019-11-09
Update favorite assets with correct count of favorites_count

### LINKED ISSUE #809: Favorite counts are off
https://alonetone.com/obbster/tracks/main-title

Count says 175 but only 2 favorites.

Migration needed to rewrite the counters on all tracks.

innatewonderer:
Fixes #809 

Updating only the assets where favorites_count did not match the count of `tracks.favorites.count` to limit unnecessary updates.


db/migrate/20191109022931_update_favorites_on_tracks.rb
```diff
@@ -0,0 +1,12 @@
+class UpdateFavoritesOnTracks < ActiveRecord::Migration[6.0]
+  def change
+    asset_ids = Track.favorites.map(&:asset_id).uniq
```
sudara: This will instantiate 45k records. I recommend using `pluck(:asset_id)` vs `map` to only select the field we need.

db/migrate/20191109022931_update_favorites_on_tracks.rb
```diff
@@ -0,0 +1,12 @@
...
+    puts "Found #{asset_ids.length} favorite assets."
+
+    Asset.where(id: asset_ids).find_each do |asset|
+      if asset.favorites_count != asset.tracks.favorites.count
+        asset.favorites_count = asset.tracks.favorites.count
```
sudara: I don't think this will actually update the database. We probably want to use `Asset.reset_counters` here https://api.rubyonrails.org/classes/ActiveRecord/CounterCache/ClassMethods.html#method-i-reset_counters

sudara:
It seems like we should think a bit about why the counts skewed. This seems like a good opportunity to use production data to get a hint. It looks like most actual counts are a few less than they should be and in a few cases hundreds less! 

```
irb(main):005:0> Asset.where('favorites_count > 10').each do |a|
irb(main):006:1* puts "favorites_count: #{a.favorites_count} actual count: #{a.tracks.favorites.count}"
irb(main):007:1> end

favorites_count: 11 actual count: 9
favorites_count: 14 actual count: 12
favorites_count: 12 actual count: 12
favorites_count: 12 actual count: 10
favorites_count: 15 actual count: 13
favorites_count: 13 actual count: 10
favorites_count: 162 actual count: 1
favorites_count: 191 actual count: 11
favorites_count: 174 actual count: 2
favorites_count: 14 actual count: 10
favorites_count: 20 actual count: 3
favorites_count: 13 actual count: 10
favorites_count: 11 actual count: 9
favorites_count: 12 actual count: 8
favorites_count: 11 actual count: 11
favorites_count: 101 actual count: 16
favorites_count: 11 actual count: 6
favorites_count: 14 actual count: 8
favorites_count: 20 actual count: 16
favorites_count: 16 actual count: 13
```

I think there are probably several reasons for this. 

1) The code is using a "toggling" pattern to manually juggle the favorites count https://github.com/sudara/alonetone/blob/ab54eb11514b7dd6fd242142e63442dda9de4d46/app/models/user.rb#L234-L242

2) Because we aren't using a counter_cache, we are probably missing out on decrementing the counter when users / tracks are being deleted. This is complicated by `SoftDeletion`, since we are moving away from using `dependant: destroy`.

It seems like we should try and setup these counters to behave a bit more like "normal" rails while we are here, to attempt to prevent them from drifting again.

db/migrate/20191109022931_update_favorites_on_tracks.rb
innatewonderer: Awesome.
Looks like `reset_counters` can be used on associations, while our favorites_count is a "custom" counter (I'm not sure that's how you classify it (we actually call `increment_` and `decrement_` counter to count those).
```
2.6.3 :006 > Asset.reset_counters(14, :favorites_count)
  Asset Load (1.0ms)  SELECT `assets`.* FROM `assets` WHERE `assets`.`deleted_at` IS NULL AND `assets`.`id` = 14 LIMIT 1
Traceback (most recent call last):
        2: from (irb):6
        1: from (irb):6:in `rescue in irb_binding'
ArgumentError ('Asset' has no association called 'favorites_count')
```
But we can use `update_counter` instead and pass it the value. Works great.

innatewonderer:
> It seems like we should try and setup these counters to behave a bit more like "normal" rails while we are here, to attempt to prevent them from drifting again.

Are you suggesting instead to have two separate `add_to_favorites` and `remove_from_favorites` methods instead of toggle? 

What I'm thinking. Let's say user B favorites an asset by user A. UserB deletes an account (we soft delete and then delete the user and its associations). However... does it REALLY mean an asset now has incorrect number of `favorites_count`? If someone ever favorited my track, shouldn't the counter stay even if user is no longer there? It's a messy code logic, but human logic - I think it makes sense. Imagine `/sudara` ever wants to remove his account. Imagine how many artists' stats are going to go down immediately. 

Thoughts?

db/migrate/20191109022931_update_favorites_on_tracks.rb
sudara: The fact that it's a "custom counter" is what triggered my main PR comment about probably needing to move away from that towards normal rails counters.

sudara:
> Are you suggesting instead to have two separate add_to_favorites and remove_from_favorites methods instead of toggle?

Yes, removing the toggle is one task, as it's a very ambiguous way of managing a count and likely how we got into trouble with the tracks with 100s of extra faves.

The other is setting up the counters like one would normally on rails, on an association in `Track`, and letting rails handle the incrementing/decrementing via a counter_cache. 

> If someone ever favorited my track, shouldn't the counter stay even if user is no longer there? 

This is a good question! If a user decides to unfavorite the track or leave the platform, we should adjust the number. The users who favorited a track are shown in the UI, so if 2 users drop off that list, I would expect to see the counter go down to match the list. This does mean that the number goes down for people, but it also means we have an accurate number and can reliably recount it in the future, etc. 

I would feel differently if we were counting favorites from guests and the number was more "total people who ever clicked like, more the merrier!" rather than "actual users on the platform who like this and you can go check out who they are."

I don't think there's a "right" answer here. It's just cleaner, easier to maintain and doesn't create dissonance in the UI to have the number match the displayed real users.

With the counter setup in a more standard way, when users are soft deleted (and their favorites are soft deleted) we should see the favorites count dip on tracks. That should be "shown" via specs, I think.

app/controllers/application_controller.rb
```diff
@@ -77,6 +77,11 @@ def find_published_asset
     track_not_found unless @asset.published? || current_user_is_admin_or_owner?(@asset.user)
   end
 
+  def track_not_found
```
innatewonderer: Moved this method from assets_controller.rb since it's only used by two methods above.

app/controllers/favorites_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+# user must be logged_in
+# when favorite track is created it triggers favorites_count counter cache on asset
+class FavoritesController < ApplicationController
+  before_action :find_user
+  before_action :find_asset
+  before_action :require_login
+  before_action :find_favorite_track, only: :destroy
+
+  def create
+    raise ArgumentError unless @asset
```
innatewonderer: This was legacy from `toggle_favorites` controller method. Not sure if it'd still be a problem for this implementation, but wanted to catch any in Bugsnag for investigation.
```
    return false unless logged_in? && asset # no bullshit	
```

config/routes.rb
```diff
@@ -115,7 +113,7 @@
       end
       resources :comments,  only: [:create, :destroy]
     end
-    resources :favorites
```
innatewonderer: we had this route as legacy. So I'm just reusing it.

app/controllers/favorites_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def destroy
+    @favorite_track.destroy
+    head :ok
+  end
+
+  private
+
+  def find_favorite_track
+    @favorite_track = Track.find_by_id(params[:id])
```
sudara: We probably want to scope through user here, to make sure user's can't delete any track id.

app/controllers/favorites_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def create
+    raise ArgumentError unless @asset
+
+    @user.tracks.favorites.create(asset_id: @asset.id)
+    head :ok
+  end
+
+  def destroy
+    @favorite_track.destroy
```
sudara: Does this make sense? Probably not....

```suggestion
    @favorite_track.soft_delete
```

app/controllers/users_controller.rb
```diff
@@ -90,14 +90,6 @@ def update
...
   end
 
-  def toggle_favorite
-    asset = Asset.published.find(params[:asset_id])
-    return false unless logged_in? && asset # no bullshit
-
-    current_user.toggle_favorite(asset)
-    head :ok
-  end
-
```
sudara: 🥇

Issue #890, opened 2019-11-04
Module::DelegationError in CreateAudioFeatureJob@default


sudara:
## Error in alonetone

**Module::DelegationError** in **CreateAudioFeatureJob@default**
open delegated to attachment, but attachment is nil

[View on Bugsnag](https://app.bugsnag.com/alonetone/alonetone/errors/5d9898864215a7001bf3e38e?event_id=5dc0360300524a6b6ec70000&i=gh&m=ci)

## Stacktrace

    app/models/asset/waveform.rb:6 - import_waveform
    app/jobs/create_audio_feature_job.rb:7 - perform

[View full stacktrace](https://app.bugsnag.com/alonetone/alonetone/errors/5d9898864215a7001bf3e38e?event_id=5dc0360300524a6b6ec70000&i=gh&m=ci)

*Created by Jenya Zueva via Bugsnag*

sudara:
An error linked to this issue has been reopened in Bugsnag
[**Module::DelegationError** in **CreateAudioFeatureJob@default**](https://app.bugsnag.com/alonetone/alonetone/errors/5d9898864215a7001bf3e38e?event_id=5dd1773e00533b158a820000&i=gh&m=ra)

sudara:
This issue has been unlinked from an error in Bugsnag
[**Module::DelegationError** in **CreateAudioFeatureJob@default**](https://app.bugsnag.com/alonetone/alonetone/errors/5d9898864215a7001bf3e38e?event_id=5dd1776500532d67f10a0000&i=gh&m=ul)

sudara:
Dupe of #916

Issue #837, opened 2019-10-08
Improve DB seeds


sudara:
## Approach

* [ ] Create helpers to create users/tracks/playlists/listens/comments in batches, relying on faker to populate 
* [ ] Have the main seeds file stay high level, mainly configuring and calling helpers
* [ ] Have a seeds file named after each edge case user, so it's easy to see where to edit and modidfy

## General needs

* [ ] Users with long names and long single words within the name
* [ ] Tracks/Playlists with long names and long single words within the name
* [ ] Users should have various states of profile filled out (no pic / with pic / no bio / with bio / no social links / all the social links)
* [ ] Enough users where at least a couple hit pagination: http://localhost:3000/users/by/last_uploaded 
* [ ] Enough playlists where this fleshes out into a grid http://localhost:3000/playlists
* [ ] Playlists with Externals Links
* [ ] Comments of various length from guest and users, some marked as spam.
* [ ] Following relationships among users
* [ ] Track descriptions and Forum posts using all markdown
* [ ] **Forum Seeds** (see https://github.com/thredded/thredded/blob/master/lib/thredded/database_seeder.rb) 

## Specific edge cases 

* [ ] Track with lyrics, images, videos in them.
* [ ] Track with no recent listeners / 5 recent listeners / 50 recent listers 
* [ ] Track with 0 favorites / 5 favorites / 50 favorites
* [ ] Track with 0 comments / 5 comments / 50 comments
* [ ] Playlist with 0 tracks and private / 10 tracks / 50 tracks
* [ ] User with 0 tracks / 5 tracks / 50 tracks / 500 tracks
* [ ] User with 0 comments / 5 comments / 50 comments
* [ ] User who has deleted some of their tracks (which also have comments and listens)

Still open questions

* [ ] Images.... Will we need a batch of 100 or something to play with?

~Open~ Old questions:
<details>
1) seeds.rb is already 300 lines and takes a while to parse — how maintainable is it to keep using 1 file? What options do we have to introduce a bit of structure there?

2) Is there a route we could take that would be less brittle? I'm imagining adding tables and relations down the line and having to comb through the seeds file and make sure each user has a value for a new col / relation. 

3) From a UI standpoint, the best way to make seeds is to evaluate the needs from the UI point of view and strategically add them, as we dev, on a case by case basis. It would therefore be best if we could devise a straightforward way to open this up so @ofsound would feel comfy adding new records, knowing they will persist. 

[...]

PR #896, opened 2019-11-09
Use correct a download syntax to download files.

### LINKED ISSUE #887: Clicking download button should download and not open in browser.
See https://alonetone.com/discuss/posts/488

`Content-Disposition: attachment;` is our friend

but this is a modern way to accomplish:

`<a download="blah.mp3" href="aaoetuh/aoeuth">some link</a>`

Just not sure if that works with redirects....

innatewonderer:
Fixes #887 

I'm seeing `filename delegated to attachment, but attachment is nil` on assets/show.
Action_Controller__Exception_caught

* [ ] All tests green


innatewonderer:
@sudara Am I correct in assuming we want the `mp3_file_name` to be provided as download name  Per comment [here](https://github.com/sudara/alonetone/blob/master/app/models/asset.rb#L73) it's been depricated? So we cannot rely on that column at all? (any plans on removing those columns in the future, or do we still need them?)
```
  # @deprecated Please use asset.audio_file.filename.
  def mp3_file_name
    if filename = audio_file&.filename
      filename.to_s
    end
  end
```
I'm still researching on a way to work with `has_one_attached :audio_file` locally if the file is missing. If you have any ideas, it'd be greatly appreciated.

sudara:
@innatewonderer Have you migrated to using the seeds yet? Manfred spent a lot of work getting everything very shiny on the new seeds, including support for downloads, etc. There's more work to be done on #837 

#890 is the same error as you seem to be getting locally. It's on Manfred's radar for next week.  

Are you constantly seeing that error on all assets? Does the download attribute work btw?

sudara:
More info on seeds just in case: https://alonetone.slack.com/archives/C03E2AE6J/p1569946263009700

sudara:
#898 and #899 were merged which may alleviate your error. But refreshing the alonetone.yml and using the seeds is the endgame...it's possible whatever your local config is pointing to just isn't there anymore.

innatewonderer:
I did update my `alonetone.yml` based on that original comment but keep forgetting to migrate to seeds (granted, it has been useful in my recent work on figuring out counter discrepancies :) But this kind of work is def more clear with seeds).

innatewonderer:
> Does the download attribute work btw?

It works beautifully.

innatewonderer:
> #890 is the same error as you seem to be getting locally — It's on Manfred's radar for next week. Are you constantly seeing that error on all assets? 

It does NOT fail when used with seed data.
It does, however, occur when running with snapshot data on all records that I didn't actually upload manually. So if there is no actual file in storage associated with the record. (I do often upload files locally to test various functionality, so those assets work just fine with new download attr).

sudara:
> with snapshot data on all records that I didn't actually upload manually. 

That makes sense, as you don't have access to production s3 keys I guess. 

Glad you got seeds sorted!

One question: what browsers did you test the download attribute on?

PR #899, opened 2019-11-11
Skip validation on unattached blob.

Manfred:
A record can have an Active Storage attachment that (no longer) points to a blob. When we attempt to validate the content-type and size of the blob we get a delegation error.

The solution in this PR is to skip validation when the attachment is not attached.

This may solve #890 and other upload related spec suite failures.


sudara:
Thank you!!! Really appreciated.

PR #901, opened 2019-11-11
Replace has_permalink with a few lines of slug code

### LINKED ISSUE #793: Rewrite Permalink logic
There are 2 bugs right now with how we use `has_permalink`.

1) I uploaded a track and https://alonetone.com/sudara/tracks/shape-361 was the URL (should have been https://alonetone.com/sudara/tracks/shape)

2) #660 Playlists can have the same permalink

This is the feature set we want for assets and playlists:

* [x] On create of an asset, Permalinks are automatically created based off ID3 tag title of the mp3 uploaded, for example using https://apidock.com/rails/ActiveSupport/Inflector/parameterize
* [x] When an asset (or permalink) is edited, *if and only if* the title changes, the permalink is updated — it's important to have good spec coverage around this, we don't want to break permalinks to older assets (even if they are unideal) unless the title changes.
* [x] The permalink is uniquely scoped to the user. If it's *not* unique to the user, a number is added to the permalink.
* [x] If `my-permalink` and `my-permalink-2` both exist, then `my-permalink-3` is used — meaning, we know how to properly increment that number and we don't have `my-permalink-1`
* [x] We need production migration to disentangle playlists where we have the same permalink used for multiple playlists.

### LINKED ISSUE #846: When reading ID3 tags from mp3, casing is not preserved / title case
"A Song Like This" should be title cased by default — or better, respect the casing in the songs ID3 tag.

Manfred:
Replace `has_permalink` in `Asset`, `Group`, and `Playlist` models.

Closes #793. Closes #846

### DIFF
--- a/lib/slug.rb
+++ b/lib/slug.rb
@@ -2,13 +2,10 @@
 
 # Helper class to generate slugs out of names or titles.
 class Slug
-  CONTROL_CHARACTERS_RE = /[\x00-\x1F\x7F]/
+  CONTROL_CHARACTERS_RE = /[\x00-\x1F\x7F]/.freeze
 
   def self.generate(string)
-    raise(
-      ArgumentError,
-      "Can't generate a slug from a nil value."
-    ) if string.nil?
+    raise(ArgumentError, "Can't generate a slug from a nil value.") if string.nil?
 
     string.unicode_normalize.downcase(:fold)
           .gsub(CONTROL_CHARACTERS_RE, '')
@@ -17,7 +14,7 @@ class Slug
           .gsub(%r{[[:space:]]+}, '-')
   end
 
-  DEDUPED_RE = %r{\A(.*)-(\d+)\Z}
+  DEDUPED_RE = %r{\A(.*)-(\d+)\Z}.freeze
 
   def self.increment(slug)
     if match = DEDUPED_RE.match(slug)
--- a/lib/slug.rb
+++ b/lib/slug.rb
@@ -11,10 +11,10 @@ class Slug
     ) if string.nil?
 
     string.unicode_normalize.downcase(:fold)
-      .gsub(CONTROL_CHARACTERS_RE, '')
-      .gsub(%r{\A([[:space:]]+)}, '')
-      .gsub(%r{([[:space:]]+)\Z}, '')
-      .gsub(%r{[[:space:]]+}, '-')
+          .gsub(CONTROL_CHARACTERS_RE, '')
+          .gsub(%r{\A([[:space:]]+)}, '')
+          .gsub(%r{([[:space:]]+)\Z}, '')
+          .gsub(%r{[[:space:]]+}, '-')
   end
 
   DEDUPED_RE = %r{\A(.*)-(\d+)\Z}
--- a/lib/slug.rb
+++ b/lib/slug.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+# Helper class to generate slugs out of names or titles.
+class Slug
+  CONTROL_CHARACTERS_RE = /[\x00-\x1F\x7F]/
+
+  def self.generate(string)
+    raise(
+      ArgumentError,
+      "Can't generate a slug from a nil value."
+    ) if string.nil?
+
+    string.unicode_normalize.downcase(:fold).
+      gsub(CONTROL_CHARACTERS_RE, '').
+      gsub(%r{\A([[:space:]]+)}, '').
+      gsub(%r{([[:space:]]+)\Z}, '').
+      gsub(%r{[[:space:]]+}, '-')
+  end
+end
--- a/lib/slug.rb
+++ b/lib/slug.rb
@@ -10,11 +10,11 @@ class Slug
       "Can't generate a slug from a nil value."
     ) if string.nil?
 
-    string.unicode_normalize.downcase(:fold).
-      gsub(CONTROL_CHARACTERS_RE, '').
-      gsub(%r{\A([[:space:]]+)}, '').
-      gsub(%r{([[:space:]]+)\Z}, '').
-      gsub(%r{[[:space:]]+}, '-')
+    string.unicode_normalize.downcase(:fold)
+      .gsub(CONTROL_CHARACTERS_RE, '')
+      .gsub(%r{\A([[:space:]]+)}, '')
+      .gsub(%r{([[:space:]]+)\Z}, '')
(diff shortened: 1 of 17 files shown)


lib/slug.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# Helper class to generate slugs out of names or titles.
+class Slug
+  CONTROL_CHARACTERS_RE = /[\x00-\x1F\x7F]/
```
sudara: Interested in what this does...

lib/slug.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def self.generate(string)
+    raise(
+      ArgumentError,
+      "Can't generate a slug from a nil value."
+    ) if string.nil?
+
+    string.unicode_normalize.downcase(:fold).
```
sudara: Is there a reason here to use custom logic vs. rely on https://apidock.com/rails/v6.0.0/ActiveSupport/Inflector/parameterize ?

lib/slug.rb
Manfred: It matches [the ASCII control characters](https://en.wikipedia.org/wiki/Control_character), things like the C string ending zero, the audible bell, etc.

spec/lib/slug_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+    ).to eq('càfé'.unicode_normalize(:nfkc))
+  end
+
+  it 'folds ASCII characters down' do
+    expect(Slug.generate('HELLO')).to eq('hello')
+  end
+
+  it 'folds Unicode characters down' do
+    expect(Slug.generate('ĲLAND')).to eq('ĳland')
+    expect(Slug.generate('Århüs')).to eq('århüs')
```
sudara: ActiveSupport::Inflector#parameterize "transliterates" all accented characters to ASCII — do you think there's a valid reason why they do that?

spec/models/asset_spec.rb
```diff
@@ -129,7 +135,7 @@
 
     it "should handle empty name in mp3 tag" do
       asset = file_fixture_asset('japanese-characters.mp3', content_type: 'audio/mpeg')
-      expect(asset.permalink).to eq("01-oaee") # name is 01-\266Ե??\313"
+      expect(asset.permalink).to eq("01-¶ôμäèë") # name is 01-\266Ե??\313"
```
sudara: This seems a bit strange! But then again, so did the original spec, this just seems stranger :)

lib/slug.rb
Manfred: See comment below.

spec/lib/slug_spec.rb
Manfred: As an aside, that method has absolutely no place on the inflector. I tried to protest it, but I lost.

No, that was just a misunderstanding of the specs combined with bad implementations in ancient browser versions. Since RFC3986 you can use any UTF-8 character in a URL but you have to percent-encode it. This is exactly what Rails does.

A permalink database value `Ĳ` would have to be encoded as: `/someone/playlists/%C4%B2`, but a modern browser will present it as `/someone/playlists/Ĳ`.

---

Same goes for badgers.

Screenshot 2019-11-11 at 17 52 20

Screenshot 2019-11-11 at 17 53 34

spec/models/asset_spec.rb
Manfred: I removed the spec because its implementation is a duplicate now.

Manfred:
All of the CodeClimate issues don't show up in local Robocop 😢

spec/lib/slug_spec.rb
sudara: > a modern browser

Do we know how modern is modern? I'm not too worried about this, but I'm usually so cavalier that it might be nice to know what we are cutting off.

sudara: My only real concern is maintenance burden, i.e. how likely is it that I'll have to come bug Manfred because we have a custom built UTF-8 friendly permalink scheme that turns out to have different quirks in different browsers.

sudara:
> All of the CodeClimate issues don't show up in local Robocop

Opened #904 — I've been wanting to adjust a few more rules as well.

sudara:
QA'd this locally and it's working like butter! Exciting.

One potential minor regression from production behavior:

When uploading a track, we take the ID3 tag and populate the title attribute. When the track finishes uploading, it goes to an edit page which is technically optional. On master, the permalink is already made from the title. On this branch, it looks like we get the default `untitled` instead of the title attribute (callback order?) so the permalink gets set to:

http://localhost:3000/carole/tracks/untitled/edit

And only changes if there is a save on the edit form.

Manfred:
> we get the default untitled instead of the title 

Ah, good catch. We can solve that.

spec/lib/slug_spec.rb
Manfred: It's hard to find because nobody seems to care much about this, but my guess is Internet Explorer 7, Safari 5, etc.

Manfred:
@sudara I can't replicate this issue, you're talking about [posting to the assets controller](https://github.com/sudara/alonetone/blob/master/spec/request/assets_controller_spec.rb#L234) right?

[Upload audio.mov.zip](https://github.com/sudara/alonetone/files/3835115/Upload.audio.mov.zip)

sudara:
@Manfred Yeah, uploading a single mp3 via the UI in Safari 13.0.3 on this branch with `storage_service: filesystem`

Interesting, didn't happen to me when I tried again. Is it possible there's a sort of race condition?

Manfred:
It's possible that the development server didn't reload entirely after switching branches or something? I don't know.

sudara:
No, I had killed the server and was poking around on the branch for a while playing with permalinks in general.

sudara:
Ok, I can reproduce with this mp3 https://www.dropbox.com/s/wmkgqhe90x2e8op/0p1coast%20live%20at%20melatonin.mp3?dl=0

image

Was wondering if it might be related to #846

Manfred:
Ah, now I understand what you mean. There is a regression where we're not falling back to the filename when creating an Asset for an MP3 file without a title in the ID3 tags. I will fix this.

spec/lib/slug_spec.rb
sudara: Ok, that's what I suspected. Sounds fine. Only other discrepancy to note for the future is that what we accept and display as a "login" might differ from asset/playlist permalinks.

Issue #660, opened 2019-08-13
Playlist permalinks: Two playlists can have same permalink, making editing impossible


ofsound:

sudara:
Good call. Is this only via uploading a zip? Or can you reproduce with manual playlist creation?

ofsound:
I just tried it by manually creating a "Test Playlist" twice, and they appear separate but overlap in editing.

sudara:
#793 describes the desired permalink behavior.

Because permalink/uniqueness implementation differs on assets vs. playlists — playlists are  suffering uniquely — there might be mulitple playlists with the same permalink.

Therefore in addition to relying on new permalink behavior in #793 we require a migration that looks for playlists with the same user and same permalink and creates new permalinks for them.

Manfred:
We might want to add a unique constraint on `user_id` + `permalink` on `playlists` and `assets`,  and on `permalink` for `groups`. For bonus points we can catch `ActiveRecord::RecordNotUnique` in the controllers and render the form when a race condition triggers the problem.

Manfred:
Did we solve this with the migration this week?

sudara:
Yes!

PR #902, opened 2019-11-11
De-duplicate playlist permalinks

Manfred:
Adds migration that loops through all playlists and de-dupes their permalink without relying on the model code.

References #660.

### DIFF
--- a/db/migrate/20191111153306_make_playlist_permalinks_unique_within_scope.rb
+++ b/db/migrate/20191111153306_make_playlist_permalinks_unique_within_scope.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'slug'
+
+class Playlist < ActiveRecord::Base
+end
+
+class MakePlaylistPermalinksUniqueWithinScope < ActiveRecord::Migration[6.0]
+  def up
+    Playlist.find_each do |playlist|
+      playlist.reload
+
+      duplicates = Playlist
+                   .where(permalink: playlist.permalink)
+                   .where(user_id: playlist.user_id)
+                   .where.not(id: playlist.id)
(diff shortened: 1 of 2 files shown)


db/migrate/20191111153306_make_playlist_permalinks_unique_within_scope.rb
```diff
@@ -0,0 +1,29 @@
...
+
+require 'slug'
+
+class Playlist < ActiveRecord::Base
+end
+
+class MakePlaylistPermalinksUniqueWithinScope < ActiveRecord::Migration[6.0]
+  def up
+    Playlist.find_each do |playlist|
+      playlist.reload
```
sudara: Not blocking, but curious why the AR object needs reloading right out of the gate...

db/migrate/20191111153306_make_playlist_permalinks_unique_within_scope.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      duplicates = Playlist
+                   .where(permalink: playlist.permalink)
+                   .where(user_id: playlist.user_id)
+                   .where.not(id: playlist.id)
+                   .order(id: :asc).to_a
+      next if duplicates.empty?
```
sudara: Thanks, very nice!

db/migrate/20191111153306_make_playlist_permalinks_unique_within_scope.rb
Manfred: Because `find_each` loads a batch and the records in the batch can be stale. It's not very likely because the records are returned in sequence and changes are always made to records with a higher ID. I thought it would be better to be safe than sorry.

PR #913, opened 2019-11-13
Change storage URL expiration from 5 minutes to 6 hours

### LINKED ISSUE #869: Image thumbnails intermittently fail to load on localhost / in percy
I believe this can be reproduced by loading the home page, blowing away 'tmp/cache' and loading the homepage again.

Maybe the underlying reason is that variant urls aren't staying consistent across page loads / server boots? 

Keep in mind we also return `304 Not Modified` responses for the home page (soon to be other locations), so perhaps the browser is using a response that points to invalid variants.

On localhost:

image

On Percy:

Build #1638 - alonetone - Percy | Visual testing as a service 2019-10-25 22-10-01

Manfred:
In development and without the Fastly mock all image variants URLs will be Active Storage service URLs. Just like other presigned URLs these URLs expire. Combined with HTML page caching this can lead to URLs expiring before the page cache expires, this leads to missing images.

This PR changes the expiration time from 5 minutes to 6 hours to make this happen less. We can't fix it completely because the URLs require an expiration time and some HTML is cached indefinitely when nothing changes to the database.

Closes #869.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -33,6 +33,10 @@ module Alonetone
     config.active_storage.service = config.alonetone.storage_service
     config.active_storage.variant_processor = :vips
 
+    # Expiration for storage URLs should be at least as long as the maximum
+    # cache time for HTML pages. The default is 5 minutes.
+    config.active_storage.service_urls_expire_in = 6.hours
+
     # Turn off previewers and analysers because we don't use them.
     config.active_storage.analyzers = []
     config.active_storage.previewers = []


sudara:
Couple questions here:

1) Do you think it's worth it to manually expire the conditional gets on the home page to match? I doubt production is lasting too long anyway, 6 hours would be reasonable. 

2) If not, can we change this local setting to 72 hours? That would eliminate seeing the expired images when we go to boot up the server the next day.

Not sure if there's another way we could link the caching and image expiration...

Manfred:
I don't think we use presigned URLs of any kind in production except maybe the download URL somewhere so expiration is not a problem there. I wouldn't change production page cache expiration just for this.

>  can we change this local setting to 72 hours?

Yes, but you I believe you will have to manually clear memcached at some point anyway. A more sustainable solution might be to implement the fastly mock into a development controller and switch everyone there.

sudara:
Ok thanks for this!!!

PR #915, opened 2019-11-13
Don't stub current user and user session accessors in spec suite

Manfred:
By stubbing `current_user_session` and `current_user` on the controller object you lose the difference between the controller instance variable and the example local variable space. Does that sentence make sense?

* Remove duplicate inclusion of `Authlogic::TestCase` in controller and request specs.
* No longer includes `Authlogic::TestCase` in model and other specs.
* Allows `UserSession` and the `AuthLogic` controller integration to work as intended.
* Adds `alphabetic` rspec order: `rspec --order alphabet`.
* Stop activating Authlogic test integration all over the spec suite.

### DIFF
--- a/spec/controllers/notifications_controller_spec.rb
+++ b/spec/controllers/notifications_controller_spec.rb
@@ -1,19 +1,23 @@
 require "rails_helper"
 
 RSpec.describe NotificationsController, type: :controller do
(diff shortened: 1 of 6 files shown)


sudara:
> Does that sentence make sense?

100% 

Thank you for this long needed cleanup!

PR #919, opened 2019-11-14
Start removing reliance on Authlogic from the code base

Manfred:
* Moves authentication and authorization helpers to a concern
* Remove direct reference to `Authlogic` from as many places as possible

References #920.


sudara:
Excited for this!

PR #941, opened 2019-11-25
Check if audio file is attached to asset

### LINKED ISSUE #916: filename delegated to attachment, but attachment is nil
## Error in alonetone

**ActionView::Template::Error** in **assets#show**
filename delegated to attachment, but attachment is nil

[View on Bugsnag](https://app.bugsnag.com/alonetone/alonetone/errors/5dcc37c11ef06d0019f435ed?event_id=5dcc59fb005331eaa0750000&i=gh&m=ci)

## Stacktrace

    app/views/shared/_asset.html.erb:29 - _app_views_shared__asset_html_erb___3028549533620303048_70233455473200
    app/views/assets/show.html.erb:1 - _app_views_assets_show_html_erb___1552870722523836647_70233455397720

[View full stacktrace](https://app.bugsnag.com/alonetone/alonetone/errors/5dcc37c11ef06d0019f435ed?event_id=5dcc59fb005331eaa0750000&i=gh&m=ci)

*Created by Jenya Zueva via Bugsnag*

Manfred:
Change the asset partial to check for the existence of an attachment before calling any method on `audio_file`.

Closes #916.

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -56,6 +56,7 @@ class AssetsController < ApplicationController
   def show
     respond_to do |format|
       format.html do
+        check_for_missing_attachment
         lazily_create_waveform_if_needed
         @assets = [@asset]
         set_related_show_variables
@@ -103,6 +104,7 @@ class AssetsController < ApplicationController
   end
 
   def edit
+    check_for_missing_attachment
     @allow_reupload = true
   end
 
@@ -205,6 +207,13 @@ class AssetsController < ApplicationController
       :user_id,
       :youtube_embed
     )
+  end
+
+  def check_for_missing_attachment
+    unless @asset.audio_file.attached?
+      flash.now[:error] = "Uh oh, there's no longer an mp3 attached to this track.<br/>
+        Please edit it and click 'upload a new version' to add the mp3"
+    end
   end
 
   def track_not_found
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -99,7 +99,7 @@ class Asset < ApplicationRecord
   end
 
   def self.latest(limit = 10)
-    with_preloads.limit(limit).order('assets.id DESC')
+    with_preloads.limit(limit).order('assets.created_at DESC')
   end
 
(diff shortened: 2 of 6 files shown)


Manfred:
Maybe we want to show a warning that the audio file is missing and instructions to the owner on how to replace it?

sudara:
I'm going to 

1) Rename the track to make it extremely obvious it's missing an mp3

2) Remove it from the playlist, as we test adding/removing/sorting, etc on that playlist (Maybe this was a bad idea and we can create a new playlist specifically for sorting testis)

Agreed about the UI need. I will add a temporary flash message to the owner/moderator and we can consider it when next working on track/playlist design.

app/controllers/assets_controller.rb
```diff
@@ -207,6 +209,13 @@ def asset_params
     )
   end
 
+  def check_for_missing_attachment
+    unless @asset.audio_file.attached?
+      flash.now[:error] = "Uh oh, there's no longer an mp3 attached to this track.<br/>
+        Please edit it and click 'upload a new version' to add the mp3"
```
Manfred: It's really weird to use flash for this because you only really need the session when redirecting.

sudara: Do you mean in general you don't like `flash.now`? We have a UI pattern around flash that makes it trivial to display messages consistently, however we are slowly moving to a different UI that is js based, since many of our pages are turbolinks/ajaxy.

The ideal UI I'd like for this particular case is to modify the playlist player and overlay a message for both users/owners, disabling the play button.

Manfred: Oops, I misread that as using the regular flash.

And yes, I don't like flash or the Android toaster because I end up missing important information. But maybe that's a personal preference.

sudara:
I’m also not a fan of Rails flash in general, mainly because it tries and
fails to be a reliable way to send messages to the UI, but implementation
depends on request specific implementation (redirect/render)

This reminds me though that I have to move a bunch of current flashes to
.now or the new js variant we have in flash_controller.

Issue #927, opened 2019-11-14
Implement Forum Illustrations


sudara:
As seen in figma: https://www.figma.com/file/YdjrVsNumbBsWVo82Wje2h6N/alonetone-white-theme?node-id=1358%3A3289

Source files:

https://www.dropbox.com/sh/lg2u3vge5xkgu85/AABglNVbEmwxXqGibee2wCQfa?dl=0
https://www.dropbox.com/s/6j7jjad1hfyv3ha/TestGnomes.zip?dl=0

Alec Instructions:

> Convert to grayscale (Image > Mode > Grayscale) and then resize to whatever dimensions you need for the site.

Samo scheduled the images as appearing 96x96 in the index view and 120x120 in the individual forum view. We should just go with 120 (so 240x240) in both a) black on white and b) inverted and resize down to on the index. 

Could probably go with png8 or grayscale png I guess?

PR #943, opened 2019-11-29
New layout and themes for thredded forum

### LINKED ISSUE #766: Implement design of forums in figma

ofsound:
Fixes #766, #927 

Sudara
* [x] Refactor SCSS theming to integrate with site toggle
* [x] Remove components.scss 
* [x] Make illustrations themable in rails
* [ ] Switch between dark/light theme images in stimulus controller

To Discuss
* [ ] breadcrumbs don't show the group, just all messageboard
* [x] Where do topic images come from
* [ ] notification settings, important enough to be top left, it's more like general settings sometimes?
* [ ] other button locations
* [x] can get stuck in new message board, or new messageboard group page without a way back
* [ ] private topics?
* [ ] read and unread icons show up where (following also a css option)
* [x] how to populate sub_nav
* [x] stimulus controller for new topics link
* [ ] get enough posts to work on pagination
* [ ] Figma for new topic popup, topic page, various button hover/active states?

TODO
* [x] spacing between inputs and their submit buttons

* [ ] All tests green


sudara:
> Where do topic images come from

Yeah, let's chat about this. I have a batch of alec images, we should probably process them the same way across the site and I need to get a few in for blank slates too.

sudara:
> how to populate sub_nav

This is probably best done manually with a new html partial. I'm not sure it's worth it to do it programatically.

sudara:
@ofsound Can you merge master in? Looks like there are some forms/white.scss stuff i'm a bit worried about screwing up.

ofsound:
> @ofsound Can you merge master in? Looks like there are some forms/white.scss stuff i'm a bit worried about screwing up.

@sudara I've done it!

sudara:
While reorganizing scss structure: 

It looks like sass `@import` is deprecated: https://sass-lang.com/documentation/at-rules/import

sudara:
> Because @use adds namespaces to member names, it’s safe to choose very simple names like $radius or $width when writing a stylesheet. This is different from the old @import rule, which encouraged that users write long names like $mat-corner-radius to avoid conflicts with other libraries, and it helps keep your stylesheets clear and easy to read!

sudara:
I've gone ahead and bundled the forum css into white/dark themes.

You should be able to move the theme specific variables into the theme.

I've also gone ahead and removed components.scss (which manually named every file to be imported) in favor of glob importing. This should help prevent any accidental or implicit reliance on order dependencies and generally be more straightforward. 

I'll get the subnav in, and then this one is waiting for the illustrations from #950 and whatever else you want to do on it. I vote we "merge early" (Ideally today!) rather than try to #pp the forum.

sudara:
I'm not sure how I feel about `@use` on first look. The pattern is familiar, like js modules. Seems like the main changes would be:

1) Whenever you want to use something from another file, it has to be explicitly referenced. It's ok to reference over and over again from many files, it'll only be included once.

2) everything you pull into a file is namespaced

3) Not sure how theming would work in it's current shape, as the theme files can't really both be `use`d — so probably we'd have to do plain css variables?

4) Overriding third party things like thredded change as the pattern of defining variables first isn't possible any more (I guess they have a solution for that).

sudara:
@ofsound There's a few strange form changes in Percy. Hoping we can wrap up a version of this today if possible, so we can get a Dark Theme beta out generally...

PR #940, opened 2019-11-22
Implement new Playlist Player and Comment styles

ofsound:
Fixes #939, #938 

To Discuss with sudara:
* [ ] Rollover states on any of these new buttons
* [x] by You by Artist logic
* [x] add to favorites animation properly removed
* [x] what to do with the delete button on track edit
* [x] how to deal with groups of comments below a playlist track
* [x] checkboxes are different sizes, site wide change?
* [x] consider new placements for edit playlist and edit descriptions buttons
* [ ] color themes for external links

## "Ready For Review" checklist

These checklists are to help ensure the code review basics are covered. Consider removing to reduce noise.

* [x] PR title accurately summarizes changes
* [ ] New tests were added for isolated methods or new endpoints
* [ ] I opened an issue for any logical followups
* [ ] If this fixes a bug, "Fixes #XXX" is either the very first or very last line of the description.

## Before code review *and after additional commits* during review.

* [ ] Update title and description to account for additional changes
* [ ] All tests green
* [ ] Migrations were tested locally and do the right thing
* [ ] Booted up the branch locally, exercised any new code
* [ ] Percy changes are purposeful or explained
* [ ] Css changes are happy on mobile (via Percy is ok)

PR #944, opened 2019-11-30
I809 add controller and js behavior to handle favorites

innatewonderer:
We want to move handling of asset favorite action from random toggle action to a RESTful.
It is also using new `Rails.ajax` methods to trigger CREATE/DELETE logic on `favorites#toggle` stumulusjs js controller action. 

The main goal for this was actually to move away from custom counters (favorites_count) to cache_counter functionality. We now achieve it by adding `has_many favorites` [association on assets](https://github.com/sudara/alonetone/pull/944/files#diff-fa23a92f3f11f10e69b71ca802d4e419R40).
`belongs_to :asset, counter_cache: :favorites_count` triggers the actual counter increase when new favorites track is created.
```
  unfavore() {
    Rails.ajax({
      url: `/favorites/delete`,
      type: 'PUT',
      data: `asset_id=${this.data.get('asset-id')}`
    })
  }
```

#### Few additional changes that had to happen for this to work:
- stumulusjs controller got upgraded with logic to handle [ajax favorites#create/delete actions](https://github.com/sudara/alonetone/pull/944/files#diff-7ba47c54ec4601c4641b7b49e345375fR33)
- change `isFavorited()` method in app/javascript/controllers/favorite_controller.js class to extract asset id from data object rather than from a URL since we no longer have an href url to be able to parse asset id out of it. 

#### Migration
Took a second to find the right method to reset the counters, apparently `update_counters` changes the counter up or down rather than recalculating the value... but `reset_counters` do just that. Tested locally. Had 7 records affected and migration cleaned it up.

* [x] All tests green
* [x] Migrations were tested locally and do the right thing
> Percy broken?


app/javascript/controllers/heart_controller.js
```diff
@@ -21,12 +22,29 @@ export default class extends Controller {
...
     } else {
       this.animation.clickFave()
       this.favorited = true
       this.faved()
+      this.favore()
     }
   }
+  unfavore() {
+    Rails.ajax({
+      url: `/${this.data.get('current-user')}/favorites/delete`,
```
sudara: This opens up an interesting Q. We have `window.username` (which should probably be renamed to `.login` and eventually we should probably move that stuff into a "real" js object) — but is there a good reason for the route to include the user at all? See follows: https://github.com/sudara/alonetone/blob/master/config/routes.rb#L45-L46

Wondering if we can just do `/favorites/234` since when we are logged in, the user is always `current_user`

Just a question of what we want to be consistent with + what's simplest...

app/views/assets/_asset.html.erb
```diff
@@ -51,10 +51,10 @@
 
 <div class="tracks_reveal" data-target="normal-playback.details">
   <% if logged_in? %>
-    <%= link_to toggle_favorite_path(asset_id: asset.id), remote: true, class: 'add_to_favorites',
-      data: {controller: 'favorite', action: 'favorite#toggle' } do %>
-      <%== render file: svg_path('svg/animation_heart.svg') %>
-    <% end %>
+    <a href="#" data-remote="true", class='add_to_favorites'
```
sudara: Just a friendly heads up that @ofsound has a branch open that includes layout/css changes to the favorite button (so you aren't surprised if yours ends up merging afterwards) https://github.com/sudara/alonetone/pull/940/files#diff-1ab2e2c4523e138fdf8a768d7f67b9caL21

app/javascript/controllers/heart_controller.js
innatewonderer: > Wondering if we can just do /favorites/234 since when we are logged in, the user is always current_user

Few reasons:
- we need a user to assign user's favorites. So it was either by ensuring we get a `current_user` correctly in the controller or by using named route
- I read it more as user's favorites `sudara/favorites/123` rather than `/favorites/123`

But I see your point though. Looked over the code I removed, and it also just says `current_user.`... Let's try the simpler route.

Would you prefer to keep `favorites#create/delete` or should I make custom names as `put :favor` and `put :favor_not` (ha LOL ... `unfavor` just doesn't sound good) ... remove_favorite and add_favorite. Names are hard.

sudara: I think I prefer favorites_controller#create/delete to favorite/unfavorite or add_favorite/remove_favorite.

app/javascript/controllers/heart_controller.js
```diff
@@ -21,12 +22,29 @@ export default class extends Controller {
...
       this.unfaved()
+      this.unfavore()
     } else {
       this.animation.clickFave()
       this.favorited = true
       this.faved()
+      this.favore()
     }
   }
+  unfavore() {
```
sudara: to use `favorite/unfavorite` as the verbs here.

sudara: Ah, looks like HeartController is defining faved/unfaved as the verbs. We could probably change those to favorite/unfavorite to be more explicit.

sudara:
Excited to get this fixed up! 

`HeartController` is extended by both `FavoriteController` and `FollowController` which share functionality, mainly around the animation and toggling state.

This means:

1. We want the new logic to go in `FavoriteController` (so we don't do the wrong thing on follows) 
2. `HeartsController` defines `fave()` and `unfave()` which is intended to be overwritten by the child controllers, so nothing new needs to be added to `toggle()` 
3. We need to QA both follows and favorites in the UI to ensure everything is happy happy.

app/javascript/controllers/favorite_controller.js
```diff
@@ -2,7 +2,7 @@ import HeartController from './heart_controller'
 
 export default class extends HeartController {
   isFavorited() {
-    this.id = parseInt(this.element.href.split('=')[1])
+    this.id = parseInt(this.data.get('asset-id'))
```
sudara: Nice! Cleaner and more explicit <3

innatewonderer:
>  extended by both FavoriteController and FollowController

Ah yes, I searched to see if that method is used anywhere, but didn't think about it being extended. thank you

spec/models/playlist_spec.rb
```diff
@@ -101,17 +101,17 @@
   context "favorites" do
     it 'should create a new playlist for a user who does not have one' do
       expect(users(:sandbags).playlists.favorites).not_to be_present
-      users(:sandbags).toggle_favorite(assets(:valid_mp3))
+      users(:sandbags).tracks.favorites.scoping { Track.create(asset_id: assets(:valid_mp3).id) }
```
innatewonderer: Take care of deprecation warning
```
DEPRECATION WARNING: Class level methods will no longer inherit scoping from `create` in Rails 6.1. To continue using the scoped relation, pass it into the block directly. To instead access the full set of models, as Rails 6.1 will, use `Track.unscoped`.
```

Took idea for this fix from https://github.com/stefankroes/ancestry/pull/442#issuecomment-530160300

innatewonderer:
Tested both `follow_` and `favorite_` controller actions. Everything looks good (JS, ActiveRecord, etc).

Including a fix for annoying DeprecationWarning.

spec/models/playlist_spec.rb
innatewonderer: The desired solution is `To instead access the full set of models, as Rails 6.1 will, use Track.unscoped`. I've tried looking through the rails issues and PRs, tried changing scope - it fixed deprecation warning but broke all tests. 

Suggestion on what this message mean and how I can fix it?

innatewonderer:
We have another deprecation warning in rspec
```
.DEPRECATION WARNING: `inline_svg` is deprecated and will be removed from inline_svg 2.0 (use `inline_svg_tag` or `inline_svg_pack_tag` instead) (called from block (2 levels) in <top (required)> at /Users/jzueva/code/alonetone/spec/request/thredded_spec.rb:12)
```
but I can't find `inline_svg` method used anywhere. Following the spec thread, it lead me to `  mount Thredded::Engine => '/discuss'` which also doesn't tell me much.
If anyone have any thoughts on where this is coming from, I could add the fix as well.

spec/models/playlist_spec.rb
sudara: It looks to me they want to explicitly disallowing "reaching through" an association to do something like .create/.new/.where if you have scopes on the first model.

Meaning `Track.create` is fine but `Track.favorites.create` has a scope before the create, which is confusing and possibly problematic (apparently).

In this case, `where(is_favorite: true)` is useful for looking up records, but isn't going to do what we might want it to do (set `is_favorite` to `true`) since it's a `where` clause for a select query. 

I'd go for the straightforward thing you want to do (can also throw out the .id calls)

```suggestion
      users(:sandbags).tracks.create(asset: assets(:valid_mp3, is_favorite: true))
```

So cahdec

innatewonderer: Well, you see? Your words make sense... why can't they say that :D

innatewonderer:
@sudara 
thoughts on deploying this? I think it should be good to go.

app/controllers/favorites_controller.rb
```diff
@@ -0,0 +1,36 @@
+# responsible for creating and deleting user's favorite tracks
+# depends on asset_id
+# user must be logged_in
+# when favorite track is created it triggers favorites_count counter cache on asset
```
sudara: Can we remove most of these comments? Some of them seem to not add anything / duplicate the code.

sudara:
Looking good! 

I trimmed down the `track_not_found` method, so looks like you'll need to merge master in and do that manually on your branch since you moved it.

Ready to merge after that + removing any extraneous comments!

app/controllers/application_controller.rb
```diff
@@ -82,24 +82,20 @@ def set_theme
...
 
-  def not_found
-    raise ActionController::RoutingError, 'User Not Found'
-  end
-
   def find_user
     login = params[:login] || params[:user_id] || params[:id]
     @user = User.where(login: login).first || (current_user if %w[new favorites].include? action_name)
-    not_found unless @user
+    user_not_found_message unless @user
```
innatewonderer: @sudara noticed new method definition here. Should I remove the old ones?

sudara: `user_not_found_message` just sets @message for the 404 page — `not_found` actually raises a routing error so we 404.... so I think this change isn't desirable.

innatewonderer: > so I think this change isn't desirable.

I don't remember what exactly I meant here :) But `not_found` is no longer defined from the looks of it.

sudara: It's in application_controller, as it's used by more than this controller.

innatewonderer: Done. Reverted this change and I see now what was happening :)

app/controllers/assets_controller.rb
```diff
@@ -213,10 +213,6 @@ def check_for_missing_attachment
     end
   end
 
-  def track_not_found
-    raise ActionController::RoutingError, 'Track Not Found'
-  end
-
```
sudara: We don't want to throw this away, it's the core of how 404s work on tracks...

app/controllers/application_controller.rb
```diff
@@ -99,12 +99,12 @@ def find_user
 
   def find_asset
     @asset = @user.assets.where(permalink: (params[:permalink] || params[:track_id] || params[:id])).first
-    @asset ||= @user.assets.where(id: params[:id]).first || track_not_found
+    @asset ||= @user.assets.where(id: params[:id]).first || track_not_found_message
```
sudara: This change and the other in the file also probably need reversion — we need to raise here, not output a string :D

sudara:
@innatewonderer Just FYI I'm going to close this one out unless you wanted to wrap it up!

innatewonderer:
I do!!! Don't close it :D 
Both of the PRs were ready for review. Maybe the diff isn't showing the comments that are out of date (?). But it should be ready.
Let me review, update master and close out comments.

PR #949, opened 2019-12-06
Implement Fastly compatible thumbnails controller for development

Manfred:
Serves thumbnails in development using an unsigned URL so it doesn't expire after a few hours of development.

Also fixes an issue when someone accidentally uploads an unusable file instead of an image for a user or playlist.

If this makes sense we can merge or replace the fake Fastly server as well.

References #295


sudara:
Wow, nice! 

Does this bypass Active Storage's implementation of thumbnailing in development in favor of a custom implementation of vips? 

If so, does that imply the cache expiration trouble we've been running into is likely a common problem with the underlying Rails' implementation?

Manfred:
> Does this bypass Active Storage's implementation of thumbnailing in development in favor of a custom implementation of vips?

Yes. I considered subclassing the internal implementation and ripping out the expiration but I thought that might be less stable in the long run.

> If so, does that imply the cache expiration trouble we've been running into is likely a common problem with the underlying Rails' implementation?

Well, it's a drawback of using presigned expiring URLs combined with caching in general. You have to use an expiring URL with an expiry time that is at least as long as the longest cache time. Rails doesn't cache anything and sets cache control to no-cache and private, so technically there is no problem out of the box.

sudara:
> Rails doesn't cache anything and sets cache control to no-cache and private, so technically there is no problem out of the box.

I've been having these issues with caching disabled locally ( no `tmp/caching-dev.txt') and can confirm the home page isn't loading as 304 Not Modified. 

Do you think turbolinks is the caching culprit?

Manfred:
I really don't know why it's happening. I guess I could investigate if you'd like.

sudara:
I appreciate the work on this branch but I think we should hold off on merging until we know more about the "why." I'm hesitant to maintain overridden framework bits without knowing exactly what's happening.

Were you able to reproduce the issue and then solve it with this branch?

Manfred:
No, I never run the app more than 6 hours at a time so I never run into the issue.

sudara:
I'm not sure the 6 hour time limit is relevant. It seems to happen just hitting the homepage within a normal session of development.

sudara:
So much good work happened here, I'm loathe to close it out. However, I'd like to investigate the root cause a bit more explicitly, so I will close it until that happens.

PR #950, opened 2019-12-10
"Gift instead of punishment" illustrations

### LINKED ISSUE #953: In addition to 404ing on a deleted track, display context
When email notifications go out, they might be clicked on after a user has deleted a track. Rather than “cold” 404ing we should give the email-clicker some context.

sudara:
Closes #953. 

TODO:

* [x] 404 basics
* [x] 500 basics
* [x] Search results not found basics
* [x] Sudara needs to add better text for 404 (user? track? was something deleted?)
* [x] 404 image needs to *always* anchor to the footer. Other images could be vertically centered.
* [x] Remove extra space on "No search results" page
* [x] Mobile friendly (I'm assuming we'll shrink the image down a bit, but they don't need to shrink all the way, we are expecting to cut probably 2/3 of the image out on 320px phones.)
* [x] Clean up padding
* [x] Do the images feel centered?
* [x] Do we need to cook up WT versions without intense colors?
* [x] Alec might help us out with smoke


sudara:
@ofsound  I've got the 500/404 in. You can visit them via `/four_oh_four` and `/error`

CSS is in illustrations.scss. I moved existing stuff there out to blank_slates.scss.

In both of these cases we are adding a body class to help get rid of padding. But could use some help on the subnav situation.... maybe it's excessive to include it.

The intense colors seem to work on dark theme, but I'm less sure on white... I have the psd with layers and can export w/o background color for the white theme to see how things look.

Whups, alonetone slipped and fell! | alonetone 2019-12-12 00-36-42

sudara:
Also, both illustrations I perceive as being a bit too much to the right. Is it possible to nudge them left a bit?

sudara:
@ofsound Might be worth it to prioritize giving this branch a bit of love first, then we can merge and use the illustrations on the forum.

sudara:
No search results example available here: http://localhost:3000/search/abcdefg%20hijklmnop

Looks like it might need some extra love since there's no subnav, which leaves a gap...

ofsound:
@sudara the display of these should be at a stable place

app/controllers/pages_controller.rb
```diff
@@ -35,13 +35,11 @@ def home; end
 
   def error
     @page_title = "Whups, alonetone slipped and fell!"
-    flash[:error] = "We have a problem. But, it is not you...it's me."
```
innatewonderer: ah but i liked this message so much :D

PR #959, opened 2019-12-13
Alternative controller implementation for authentication

Manfred:
* Uses Rails session to store the authenticated session.
* Authenticated sessions don't expire.
* Authenticated sessions work with `current_user` so they are a drop-in replacement.

This PR does not replace any of the routes or login and logout links so it doesn't actually do anything at the moment. You can sign in with the new code by going to `/authenticated_session/new`. Sign out requires a delete request, so that's not very easy at the moment.

This is mostly a suggestion to see if this direction makes any sense.

### DIFF
--- a/app/controllers/authenticated_sessions_controller.rb
+++ b/app/controllers/authenticated_sessions_controller.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+class AuthenticatedSessionsController < ApplicationController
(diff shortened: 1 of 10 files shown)


sudara:
I'm into it!

sudara:
Going to close this until there's another round desire/momentum to move forward with replacing Authlogic.

PR #960, opened 2019-12-14
Use `display name` in forum rather than `login` for forum.

### LINKED ISSUE #946: Consider showing artist display names, along with or instead of /usernames/ in the forum

innatewonderer:
Fixes #946 
Tested locally
alonetone___Welcome_to_your_messageboard_s_very_first_thread_and_jzueva_Ulybnis__ruby_

* [x] All tests green


sudara:
Hey @innatewonderer! Thanks for tackling.

The reason that logins were chosen is because of the ability to @ someone on the forum, which is one of the key features we wanted. We'd need to:

1. Confirm that ability works with display names which can a wider range of characters (vs. logins which can't have spaces or special characters).
2. Confirm that the dropdown that appears when typing @ is populated with display names and NOT logins

If those two things are true, we can merge this. 

Otherwise the Plan A was to show the display name in the forum but *also* show the login with an @ in front of it, so people would know which username they could use to reply to someone. This would require overriding the thredded view (copying it into our app) where the user is displayed.

innatewonderer:
Hmmmm does this look OK?
alonetone___Testing_new_topic

config/initializers/thredded.rb
```diff
@@ -7,6 +7,10 @@
 # must be unique.
 Thredded.user_name_column = :login
 
+# User display name method, by default thredded uses the user_name_column defined above
+# You may want to use :to_s or some more elaborate method
+Thredded.user_display_name_method = :forum_name
```
innatewonderer: I guess Thredded accounted for us cranky developers :)

sudara:
@innatewonderer What shows up in the dropdown? Is it whatever `user_display_name_method` returns or is it `@username`

app/models/user.rb
```diff
@@ -268,6 +268,10 @@ def avatar_image_location(variant:)
     )
   end
 
+  def forum_name
+    "@#{login} - #{display_name}"
```
sudara:
```suggestion
    "#{display_name} (@#{login})"
```

sudara:
Ahh, looking at the screenshot, it seems the UI uses both `user_display_name_method` and `login`? but only `login` for the mentions dropdown?

If so, then we are good to go, but I'd like to change the formatting to: `Display Name (@login)`

innatewonderer:
> What shows up in the mentions dropdown? Is it whatever user_display_name_method returns or is it :login?

@sudara not sure why, but for me, there is no dropdown even on alonetone itself so I can't test it.

innatewonderer:
Fullscreen_12_18_19__8_22_PM

sudara:
I think we are missing a bit of clarity here — or at least I am, in part due to the `moderator Moderator(@moderator)` thing in the last screenshot — is that showing the login + display name (which then shows the login again?).

If we just have the forum_name be `:display_name`, can we search for the person using their display name? 

If so, that's ideal and adding the `(@login)` to the forum_name is probably overkill/confusing.

innatewonderer:
Oh, hey, would ya look at that. We are able to show `display_name`

Fullscreen_1_31_20__6_39_PM

And I confirmed that it turns into a link when you click submit

Fullscreen_1_31_20__6_38_PM

The only issue is that it displays the name that contains a space in the middle as a string :/ a bit unappealing.

@sudara thoughts?

PR #968, opened 2019-12-19
Don't create tracks through a scoped association

Manfred:
Rails 6.1 will starting performing the equivalent of `Track.unscoped.create` when creating through an association or relation. I'm guessing they don't want to apply attributes from a scope that is really meant for queries. This solves a deprecation warning.


sudara:
I believe this is also resolved in #944 which aims to clean up some old & messy favorite/track behavior.

Manfred:
Alright, closing this.

Issue #842, opened 2019-10-09
"Gift instead of Punishment" illustrations


sudara:
I think we want to do 3-4 of these:

* No search results: "We couldn't find any search results for ABC XYZ!"
* 404 not found: "Hmm, that doesn't seem to exist" (could reuse the same illustration as above if you like, or variation)
* 500 error: "Yikes! An error! We've been notified"
* Down for maintenance (aka fail whale, see https://www.theatlantic.com/technology/archive/2015/01/the-story-behind-twitters-fail-whale/384313/ )

Regarding sizes, I'm thinking we should consider an rough area 320px by 400px to be the area that will display on all devices, including the smallest phones (which range 320-500ish wide). Any message/text/important stuff would happen in those windows.

Optionally we can fit "more" of an illustration onto mobile by shrinking the size down, maybe even just a bit. 

Do we want to use browser text here too vs. drawn?

<img width="1248" alt="8*) * 2019-10-09 19-48-11" src="https://user-images.githubusercontent.com/472/66506875-6fb7f480-eace-11e9-9619-d3ddecda4485.png">

One idea I had if you don't want to draw a mega 1500x1300 rectangle would be to actually have two illustrations, a "footer" and main illustration, like so:

<img width="1248" alt="8*) * 2019-10-09 19-48-11" src="https://user-images.githubusercontent.com/472/66507282-32a03200-eacf-11e9-87c2-af1a80432e07.png">

sudara:
@phasesevencomics Whups, meant to tag you on this.

PhaseSevenComics:
This is rad, thanks so much for all the pixel dimensions and everything. I'll get to brainstorming on these....

PhaseSevenComics:
These are all done!

sudara:
Amazing work, thank you Alec!!!

PR #975, opened 2020-01-23
Illustrations wrapup

sudara:
Sudara
* [x] Export new 500 page with smoke going out to the right
* [x] Export Island Uke light/dark for home page hero
* [x] Add home page hero html

Ribs
* [x] CSS love on http://localhost:3000/500 — On desktop, smoke should go to the end of the browser window. On mobile, the intention is to crop so we only see the guitar pedal gnome, not scale the image down. See https://github.com/sudara/alonetone/issues/842#issue-504805070
* [x] contributor-less-noise.jpg has the wrong background color on dark theme Fixed by #974
* [x] Home page hero light / dark ala https://www.figma.com/file/YdjrVsNumbBsWVo82Wje2h6N/alonetone-white-theme?node-id=1730%3A820
* [x] Make maintenance page mobile friendly. See http://localhost:3000/maint.html — the css is also inside that file — the image comes out of public/images, as it will be served by nginx, not rails (as rails will be down). The goal is to have image as large as possible. On desktop, it looks best with dark gutters. On mobile, we could make it so the full width of the image matches the width of the device, I guess (could get rid of the dark gutters).

Questions re: css that came up:

1. True or False: It seems like we are fully granular when it comes to colors — meaning, if one adds an element with a color that is different on white/dark, then a line needs to go into dark/white.scss and there is not really any reuse of colors. Some information about this should probably be documented.

2. Colors are grouped by filename dark/white.scss — is the grouping important? It doesn't seem alphabetical. Should we make it that way? The "why" is worth briefly documenting, even if it's just "We are grouping these by filename, as each variable is unique to the file its used in"

3. It looks like there's still a bit of direct color usage by files in components, such as `site_header.scss`. I'm assuming this is the last of em, though?

Basically, looking to have another conversation around moving to css vars, so just want to have my facts straight and have things be clear and clean for any refactoring.


sudara:
@ofsound There's open work for the 500, the maintenance page and the home page illustration here. I might do the additional blank slates on another PR, just so we don't have too big batches of work!

ofsound:
@sudara 

_1. True or False: It seems like we are fully granular when it comes to colors — meaning, if one adds an element with a color that is different on white/dark, then a line needs to go into dark/white.scss and there is not really any reuse of colors. Some information about this should probably be documented._

True. we are fully granular. I have one solid candidate for a shared color, so that layer may still end up re-appearing.

_2. Colors are grouped by filename dark/white.scss — is the grouping important? It doesn't seem alphabetical. Should we make it that way? The "why" is worth briefly documenting, even if it's just "We are grouping these by filename, as each variable is unique to the file its used in"_

That order was important for me to make the tranisition, certain priority for making sense of that. But at this point the file name sections could just as well be alphabetized.

_3. It looks like there's still a bit of direct color usage by files in components, such as site_header.scss. I'm assuming this is the last of em, though?_

There are a handful of stragglers here and there, perhaps a few more in shadow definitions.

Issue #1010, opened 2020-04-30
Image attachments need content type verification


sudara:
## Error in alonetone

**ActionView::Template::Error** in **users#show**
ActiveStorage::InvariableError

[View on Bugsnag](https://app.bugsnag.com/alonetone/alonetone/errors/5ea8d5b4f295590017246143?event_id=5eaab05a005b2db66add0000&i=gh&m=ci)

## Stacktrace

    app/models/image_variant.rb:32 - to_active_storage_variant
    app/models/image_variant.rb:37 - variant
    app/models/playlist.rb:148 - cover_image_location
    app/helpers/playlists_helper.rb:18 - playlist_cover_url
    app/helpers/playlists_helper.rb:25 - playlist_cover_image
    app/helpers/playlists_helper.rb:36 - playlist_cover
    app/views/shared/_playlist.html.erb:3 - block (2 levels) in _app_views_shared__playlist_html_erb___2502125241019825502_69988095750540
    app/views/shared/_playlist.html.erb:2 - block in _app_views_shared__playlist_html_erb___2502125241019825502_69988095750540
    app/helpers/record_tag_helper.rb:99 - content_tag_for_single_record
    app/helpers/record_tag_helper.rb:85 - block in content_tag_for
    app/helpers/record_tag_helper.rb:84 - map
    app/helpers/record_tag_helper.rb:84 - content_tag_for
    app/views/shared/_playlist.html.erb:1 - _app_views_shared__playlist_html_erb___2502125241019825502_69988095750540
    app/views/users/show.html.erb:24 - _app_views_users_show_html_erb___4466462974517396422_46969213492100

[View full stacktrace](https://app.bugsnag.com/alonetone/alonetone/errors/5ea8d5b4f295590017246143?event_id=5eaab05a005b2db66add0000&i=gh&m=ci)

*Created by sudara via Bugsnag*

sudara:
Closed by #1011

PR #1011, opened 2020-04-30
Image content type validations

sudara:
Currently we aren't validating image content type for user avatars or playlists like we were before the Active Storage migration.

That's leading to issues like #1010 where `ActiveStorage::InvariableError` is raised.

I'm trying to reuse the Active Storage `attached` validator used for `Asset`, however the controller seems to hit AS logic before validations have a chance to run.

There was some chatter of a similar issue being fixed in 6.0, but `ActiveStorage::InvariableError` still seems to be hit before playlist validations are hit on this branch.

https://github.com/rails/rails/issues/32449#issuecomment-404938746

Any thoughts @Manfred?


app/models/playlist.rb
```diff
@@ -34,6 +34,10 @@ class Playlist < ActiveRecord::Base
   # We have to define attachments last to make the Active Record callbacks
   # fire in the right order.
   has_one_attached :cover_image
+  validates :cover_image, attached: {
+    content_type: %w[image/png image/jpeg image/jpg image/gif],
+    byte_size: { less_than: 20.megabytes }
+  }, if: :cover_image_present?
```
sudara, line 40: Cover images are optional, but even if i remove this if, the validation doesn't seem to run before it hits an AS error.

Manfred:
I'm confused, this PR does not solve the issue? Or does it?

sudara:
@Manfred I hoped this branch would solve the issue, but the problem seems deeper, so the PR is just a way to share code and context. Do you happen to remember if content type validation was problematic in the migration or if there was a technical reason that images aren't validated?

Manfred:
The exception you are referring to is caused by the views attempting to do something with the invalid attachment. When the Playlist validation rejects the cover image it just sets an error on `:cover_image` and `save` returns false. Then the edit view attempts to render a cover image with whatever weird state `@playlist.cover_image` is in. So I guess the solution is to reset the original cover image somehow when there are errors on `:cover_image`.

PR #1083, opened 2020-12-12
Bjm front end 2020 rejuvenation part two

### LINKED ISSUE #994: Maintenance page need styles

### LINKED ISSUE #1082: Reset password form field should have a light background
Screenshot 2020-12-11 - Login | alonetone

ofsound:
Fixes #1082
Fixes #994

▹ maintenance page

Change your login [    ] allowing for way too many characters ( and add line break for mobile )

the line height in the error explanation ol, Details must be at least 40 characters: Please link to your music/website or add more info

error'd input field placeholder text contrast increased,  ( change password from edit profile )

from playlist editing error page, change label from "Back to Editing Your Profile" to "Back to Edit Profile"

fix type in Why I built alonetone, "stuff i'm working on tonight"

add style with real bold instead of pseudo bold for <strong> used on basic page ( About > Help,  'from your old email account')

* [ ] All tests green


app/assets/stylesheets/components/edit_profile.scss
```diff
@@ -581,79 +580,4 @@
       }
     }
   }
-}
-
-// LEGACY BELOW
```
sudara, line 586: Should we open an issue to remove these, or are they unused?

app/assets/stylesheets/components/forms.scss
```diff
@@ -170,6 +172,25 @@ form.thredded--form input[type="text"], form.thredded--form textarea {
   color: $forms-placeholder-text;
 }
 
+.field_with_errors input {
+ &::-webkit-input-placeholder {
```
sudara, line 176: Is there a reason we aren't using https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder ?

public/maint.html
```diff
@@ -30,51 +25,110 @@
...
+  <div id="image_container">
+    <img id="image" src="/images/502.png" alt="">
+  </div>
+  <div id="message">
     <h1>alonetone is down for maintenance</h1>
     <h2>We probably took the site down on purpose! It should be back up shortly...</h2>
   </div>
+<script>
+
+  var viewportResized = function () {
```
sudara, line 105: Thanks! We should probably stay consistent with the codebase and use `const` for variables that won't get reassigned and `let` for variables that get reassigned, otherwise perfecto, thanks for cooking this up!

sudara:
Maint page looks good, thanks for handling. I think we should make 2 visual tweaks:

1. There should always be a vertical column of light gray. 
2. If it's trivial to let the max width of the image increase, we should let it grow further.

Screenshot 2020-12-12 - alonetone is down for maintenance 2

app/assets/stylesheets/components/edit_profile.scss
ofsound: None of these classes are in use anymore. The bottom 4 were in the old edit profile page. The top #reset-password class was used in the reset your password page, but that style (dark background) is no longer needed as of this PR.

app/assets/stylesheets/components/forms.scss
ofsound: Good point, I'm gonna remove the browse specific rules.

app/assets/stylesheets/components/edit_profile.scss
sudara: Ah I think I got confused (thought they were commented out but still in the repo). Thanks for removing em!

PR #1084, opened 2020-12-12
Add Patrons and pagination to user browsing

### LINKED ISSUE #174: Improve support for marking upload tracks as spam
This page currently includes users who only have spam tracks:
https://alonetone.com/users/by/new_artists

Additionally, we display spam tracks as if they are "private" which causes some confusion for users.

So the main goal is to make spam handling for tracks a bit better.

* [ ] Right now, on `update` tracks are marked as "private" which is a misnomer. Instead, let's add a new column called `is_spam` (We can't call it `spam` because `rakismet` adds a `spam?` method which fires the API call). 
* [ ] Instead of happening on update, we should fire off a job which makes the API call to check spam status. There can be 1 job per track, or 1 job for all tracks uploaded (since folks can either upload a zip or multiple files) This is the current code that runs on update: https://github.com/sudara/alonetone/blob/master/app/controllers/assets_controller.rb#L159-L160 — either way, we need some specs that a job was created.
* [ ] We should only update `User#assets_count` after we know the track is not spam, this should have specs.
* [ ] There need to be some changes (with specs) to `AssetsController#spam` and `#unspam`: If a track is manually marked as spam (this can be done from the UI) it should reduce the `assets_count` by 1.  https://github.com/sudara/alonetone/blob/master/app/controllers/assets_controller.rb#L186-L198
* [ ] Right now when tracks go from private to public, we email out to their followers. We want to do the same thing when a track goes from spam to not spam (so false positives don't suffer when unspammed) https://github.com/sudara/alonetone/blob/master/app/models/asset.rb#L117-L123

### LINKED ISSUE #837: Improve DB seeds

### LINKED ISSUE #971: Implement remaining blank slates
See #767 for text and detail.

### LINKED ISSUE #1074: Last Uploaded only showing a few users

sudara:
Resolves #1074
Resolves #971 
Resolves #837 

TODO:

* [x] Update https://github.com/sudara/sometimes
* [x] Patron model and migration
* [x] Organize seeds
* [x] Add Patrons to seeds
* [x] Add extra users to seeds
* [x] Blank Slate css extraction
* [x] Get pagination working
* [x] Make sure we have seeds covering "musicians" and "dedicated listener" etc

Frontend
* [x] The users subnav now can be "current" as well as still being a link —  I think we need a `li.current > a` that looks the same as `li.current`? Alternatively, we could disable the link there. I just thought it might be nice/expected as a shortcut to get back to page 1.
* [ ] Look into animation jank between user page loads
* [x] Make sure a page with just a few users on it looks ok 
* [ ] Make sure blank slates look good on /upload /playlists/new and /userhome

Questions:
* [ ] should we put less user results on a page for mobile?

Extra credit:
* [ ] `page_nav_js`

QA
* [ ] Make sure visiting /users actually highlights something (or change the url)


sudara:
@ofsound Another one to check out here. The blank slate seems only happy in a `.get_an_account_container` but I'm trying to add one to the users page. I think I inadvertently used the `.get_an_account_container` class for the mass invites, which needs to change too. We also will need it blank slates working on the user home page and upload pages....

ofsound:
> @ofsound Another one to check out here. The blank slate seems only happy in a `.get_an_account_container` but I'm trying to add one to the users page. I think I inadvertently used the `.get_an_account_container` class for the mass invites, which needs to change too. We also will need it blank slates working on the user home page and upload pages....
> 
> http://localhost:3000/users/by/patrons

I split up these responsibilities, so there's no more get_an_account anything.

.page_container_with_blank_slate  is the class to modify a page container to include the fancy background.

One issue with this current implementation is that the background was designed to tuck underneath but still show around its following boxed content, on get_an_account and mass_edit. Maybe this page would call for a slight variation where the full cloud is visible and is completely above and clear from what's below it. As is, the blank slate implementation is linked to that background quirk.

Also, the sudara image, being not as tall as the ones previously used, leads to some un-ideal display states. I guess it's tempting to write a Javascript layout which can use the aspect ratio of the desired image and the total height of the text included to better account for layout across all widths. They way that currently collapses at different widths using just CSS is sorta of hanging by the thread of having the images be like ALEC and the Text Content being about that length too.

sudara:
> I split up these responsibilities, so there's no more get_an_account anything.

Awesome!

> One issue with this current implementation is that the background was designed to tuck underneath but still show around its following boxed content, on get_an_account and mass_edit. 

Can we see what it looks like with the background tucked under the user grid? Might be fine...

I've added patrons to the db seeds (and more users generally). You can

```
bundle exec rake db:reset
```

Screenshot 2021-02-12 - Patrons - Musicians and Listeners | alonetone


> Also, the sudara image, being not as tall as the ones previously used, leads to some un-ideal display states.

The image heights should be identical for all the contributor images (550px). Is it a problem that the widths are different? We could maybe make those consistent too, if so. Probably more ideal than working around with js.

Speaking of js, when switching between user pages on desktop I get flashes of (i think?) js resizing user images. I'm wondering if we can get rid of that jank somehow, maybe default to largest size?.... I'm going to add to the PR's todos...

Screenshot 2021-02-12 - New Artists - Musicians and Listeners | alonetone

sudara:
@ofsound I added a "ben" blank slate on http://localhost:3000/upload to help test the blank slate stuff. You'll have to be on an account with 0 tracks (or temporarily remove the if statement)

sudara:
@ofsound Ok, I should be done here with backend/seeds. You can run `bundle rails db:reset` again to get the updated seeds after pulling.

There are a couple open frontend todos. Happy to help on anything...

sudara:
Screenshot 2021-03-04 - Last Seen - Musicians and Listeners | alonetone


I think the users not having any min height in css might be contributing to the flicker? This is how things load if i block the js...

Specs should be fixed....

sudara:
A doozy! Nice werrrrrk. Going to merge this for now and we can followup with any details!

PR #1086, opened 2020-12-18
Add Stimulus controller to Playlist View to layout playlist credits and dynamically manage sidebar bottom right border radius

### LINKED ISSUE #704: Refine cover view radius intersections for all album track count scenarios
Might need to be solved with either Javascript or Rails setting classes

Screenshot 2019-08-28 11 38 28

ofsound:
Fixes #704

We created a stimulus playlist_controller and hooked it into the playlists/show template.

Because of the nature of flexbox managing the layouts of the sidebar and cover view across different screen widths, some javascript was needed to fix the layout of optional credits that can float below those two. The credits were absolutely positioned so they could float below the cover, but this was problematic if they were long.

This script calculates the space needed to show those correctly and adds it via padding of the playlist_and_track_content div.

The reason the padding of the main element is needed in the calculations is because the credits div is still absolutely positioned, so it's floating free of the regular HTML flow and doesn't respect that padding above the footer.

The border-radius code detects if the sidebar column is taller than the cover view, and if so, it will dynamically add a rounded edge to it's bottom right corner.

* [ ] All tests green

### DIFF
--- a/app/javascript/controllers/playlist_controller.js
+++ b/app/javascript/controllers/playlist_controller.js
@@ -1,15 +1,15 @@
 import { Controller } from 'stimulus'
 
 export default class extends Controller {
-  static targets = ['credits', 'sidebar', 'content']
+  static targets = ['credits', 'sidebar', 'content', 'cover']
 
   initialize() {
     this.resize()
   }
 
   resize() {
-    if (this.isCoverView()) {
-      if (this.hasCredits()) {
+    if (this.hasCoverTarget) {
+      if (this.hasCreditsTarget) {
         this.alignCredits()
       }
       this.roundBottomRightCorner()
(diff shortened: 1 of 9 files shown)


app/javascript/controllers/playlist_controller.js
```diff
@@ -0,0 +1,59 @@
...
+        this.sidebarTarget.classList.remove('rounded_bottom_right_corner')
+    }
+  }
+  
+  hasCredits() {
+    return this.hasCreditsTarget
+  }
+
+  isCoverView() {
+    return document.body.classList.contains('cover_view')
```
sudara, line 42: When possible, it's best to only query the DOM via stimulus targets inside the controller scope. I would:

1) make `.large_cover_and_credits` a target 
2) use the [hasABCTarget](https://stimulusjs.org/reference/targets#properties)
3) Use `this.hasCoverTarget` instead of a new method

ofsound: 👍🏻

sudara:
I'm not sure why Percy is choking on javascript (or the hover) for that screenshot — I opened a support issue with them.

My vote is we slap a max width on there and then ship it.

I also added eslint in code climate and adjusted the js here to please it — this means that from now on it'll comment in pull requests on js "style" — It might be worth making sure vs code is setup to use eslint, you can do this by adding a string literal somewhere surrounded by double quotes, it should complain.

PR #1087, opened 2020-12-29
Rails 7 upgrade with turbo

sudara:
## DEPLOYING REQUIRES CACHE CLEAR

`app/packs` is no longer a thing since Rails 7, so no moving files is necessary, everything stays in `app/javascript`

GOAL:

* Change as little behavior as possible, just deploy the update to Rails and Turbo.

REALITY:

* All of playlist playback js was heavily refactored

Player QA Issues

* [x] Play on playlist should start loading on big play
* [x] Play on big play should start playback
* [x] Play track 1, pause. Play track 2: 100% visual progress
* [x] Click on a played track for the second time sets big play timer to 0:00
* [x] Click on track that already played results in big play time @ 0:00

TASKS:

* [x] Turbolinks -> Turbo
* [x] Update webpacker to 6.x
* [x] Get rid of Rails.fire
* [x] link_to remotes need to be replaced
* [x] Update form: get an account
* [x] Update form: login
* [x] Update forms
  * [x] Get An Account
    * [x] create account after approved 
  * [x] Login 
    * [x] bio 
    * [x] profile links
    * [x] change password 
    * [x] change login 
  * [x] create playlist
  * [x] edit playlist
  * [x] upload track 
  * [x] mass invite
  * [x] password reset
    * [x] form on login page 
    * [x] post reset pick a password form
  * [x] follow button
  * [x] favorite button
* [ ] Playlist playback

Frontend:

* [x] login form styles

QA:

* [ ] Sudara: Check production bundle size 
* [ ] Toggles on edit
* [ ] Theme toggle
* [ ] Playlist ajax navigation
* [x] Playlist sorting
* [x] Check login form success and error
* [ ] Check get and account form success and error
* [ ] Check mass invite form success and error
* [ ] Other forms? 
* [x] Favorite button
* [ ] Follow button
* [ ] Playlist playback


sudara:
@ofsound I went ahead and added inline errors to the login form. 

I was wondering if you could look at the html and css in user_sessions/new.html.erb and login.html.erb and the css there to determine the best shared path for the form styles....

You'd have to yarn install for this branch and restart `bin/webpack-dev-server`, if there's problems, blow away the node_modules dir.... Additionally, be aware that javascript has now moved into `app/packs` and therefore there some 5% chance you might have to manually move MorphSVGPlugin.js into the right place? Not totally sure since git is told to ignore that file.

jlmitch5:
I checked out this branch and it seems to work great on latest node 16 (nvm install 16 / nvm use 16 ...right now it's version 16.10.0).  I'd recommend adding a note to use that version in the contrib doc.

sudara:
@ofsound Ok, this is ready for player QA on single and playlist playback.... I'm going to take a look at tests next....

ofsound:
Gannon SVG covers don't appear when returning to cover view

Multiple versions of the same track added to a playlist can't be navigated to from one another in the sidebar playlist

Using the browser Back Button can often have no effect on the site while used during or after Playlist Player

If you click back to profile from a playlist, and then return to that playlist, it briefly flashes the full state before showing the cover view:

https://user-images.githubusercontent.com/407724/149046044-9310f61c-3ef3-468f-ac84-792f0dd07b9f.mov

Handling of Playlist Player track ends, and replaying or animation states is out of sync. ( Includes play/pause states of playlist sidebar as well ):

https://user-images.githubusercontent.com/407724/149046104-cffb6b4a-44bf-4069-900c-250d111e78ae.mov

Should the to-pause animation happen every time you navigate back to viewing the currently playing track?

https://user-images.githubusercontent.com/407724/149046489-4dd8e56c-db83-48f5-bcb9-0f9228ba0e94.mov

sudara:
@ofsound All of these are resolved except the "multiple of the same track" problem. I'm going to open up an issue for that one, as it will require some backend shuffling....

deploy deploy deploy?.... wanna do one more poke around?

sudara:
Ok, I fixed the multiple tracks thing. You can't *link* to anything other than the first copy, but you can at least select the tracks in the sidebar and stitches will let you jump between them as different tracks.

Issue #1088, opened 2021-01-17
Invites part I


sudara:
A starting place to allow moderators to create invite links. (#874 goes into all hopes and dreams...)

* Moderators can generate a unique invite link that can be used by many people and knows the number of active users that were created from it
* Moderators can generate a single use invite link which gets sent to particular email address. 
* When a prospective user opens a single use invite link, the email address is pre-filled on a user creation form. 
* Users created by an invite link should be attributed to the person who invited them (see: `invited_by_id` on User)
* Moderators can list active invites and it's clear which are single use vs. open ended
* Moderators can mark invites as inactive / invalid at which point they will no longer function
* Invalid invites redirect to `/get_an_account` with a flash message apology

```
invite
    email
    token
    invited_by_id
    single_use?
    users_created_count
    valid?
    created_at
```

Notes:

* Emails are unique to users on alonetone, so validation has to happen before invite is sent.

Questions:

* Email verification normally occurs after account creation before log in is allowed. Can it be skipped if the person is coming via a one-off emailed link?
* Is there a valid reason to store the `invite_id` on the user (or some other relationship)? We already store `invited_by_id` (because of account_requests and the plan to allow user to user invites).

Manfred:
> Moderators can generate a unique invite link that can be used by many people and knows the number of active users that were created from it.

This is going to behave significantly different from a use-once `Invite` so I don't think we should merge them into one model. Can I call this a `GroupInvite`?

sudara:
@Manfred Sure! I guess to some degree `GroupInvite` really is just a token that gives access to the "create new user" page that can be revoked and tracked. If we are going to separate the model, perhaps we can also add a `name` field so the front end can say "Welcome, RPM Challengers!" or "Welcome, r/wearethemusicmakers" etc....

sudara:
Maybe `MassInvite` would be better, as a `Group` is another feature half-built, just lower priority than all the work that we have been doing.

Manfred:
Yes, I like `MassInvite`.

sudara:
Maybe this is out of scope for now, but ideally we'd be able to see the users who signed up via a mass invite link. `User.invited_by_id` is there to track moderator-approved accounts and future mod-to-user and user-to-user invites, but perhaps not "enough" for `MassInvites`

sudara:
👏🏼 👏🏼 👏🏼

Issue #732, opened 2019-09-09
Single emoji comments maybe systematically rejected as "That didn't work"?


ofsound:
<img width="339" alt="Screen Shot 2019-09-09 at 4 33 31 PM" src="https://user-images.githubusercontent.com/407724/64571006-9512e080-d31f-11e9-874a-d3db393937d5.png">

sudara:
Rails believes the comments are duplicates.

<img width="862" alt="image" src="https://user-images.githubusercontent.com/472/64732601-647ba400-d4e3-11e9-8972-11ab0e88663e.png">

sudara:
Yikes, this is pretty open

https://github.com/sudara/alonetone/blob/b55f57db2013f5c538fd74277548c81f7fc00e4c/app/models/comment.rb#L41

sudara:
It considers these two emojis equivalent: 

```ruby
irb(main):003:0> Comment.where(remote_ip: '192.168.1.107', body: '🤑')
  Comment Load (115.8ms)  SELECT `comments`.* FROM `comments` WHERE `comments`.`deleted_at` IS NULL AND `comments`.`remote_ip` = '192.168.1.107' AND `comments`.`body` = '🤑' LIMIT 11
=> #<ActiveRecord::Relation [#<Comment id: 238128, commentable_type: "Asset", commentable_id: 112990, body: "💀", created_at: "2019-09-11 20:26:51", updated_at: "2019-09-11 20:26:51", commenter_id: 1, user_id: 1157, remote_ip: "192.168.1.107", user_agent: "Mozilla/5.0 (iPhone; CPU iPhone OS 13_1 like Mac O...", referrer: "http://neve.local:3000/", is_spam: false, private: false, body_html: nil, deleted_at: nil>]>
```

sudara:
Not all of mysql on production is set to utf8mb4 :/

```sql
mysql> SHOW VARIABLES WHERE Variable_name LIKE 'character\_set\_%' OR Variable_name LIKE 'collation%';
+--------------------------+--------------------+
| Variable_name            | Value              |
+--------------------------+--------------------+
| character_set_client     | utf8               |
| character_set_connection | utf8               |
| character_set_database   | utf8mb4            |
| character_set_filesystem | binary             |
| character_set_results    | utf8               |
| character_set_server     | utf8               |
| character_set_system     | utf8               |
| collation_connection     | utf8_general_ci    |
| collation_database       | utf8mb4_unicode_ci |
[...]

PR #1096, opened 2021-01-24
Allows admins to create mass invites and visitors to use them

Manfred:
* Adds `MassInvite` for mass invites and `MassInviteSignup` to count users signups through the mass invite.
* Adds admin controllers and views to manage mass invites.
* Adds a user's origin so you can see if they were mass invites, invited, or were born from the dark dark void.
* Allow users to create an account through an active mass invite link.
* Move the user creation spam stuff into a concern in case we need it for other controllers that create accounts and maybe refactor the users controller<sup>1</sup>.

Frontend Todos

* [x] Blank slate copy
* [x] Prettier route
* [x] After user creation, make it clear user needs to check email
* [x] Password fields should be on top of each other
* [x] Check out expired invite situation / copy
* [ ] QA whole flow as a n00b guest
* [x] "show" page for the mass invite that lists users? Or subcategory for mass-invited signups under "All Users" in admin
* [x] @ofsound can look at admin form (might be unifying forms across the site)

References #1088

---

* <sup>1</sup> It looks like all the `flash` related code in the users controller doesn't do anything because it's not show on the login page.

### DIFF
--- a/app/assets/stylesheets/components/blank_slate.scss
+++ b/app/assets/stylesheets/components/blank_slate.scss
@@ -60,6 +60,10 @@
         display: none;
       }
     }
+    a {
+      text-decoration: underline;
+      color: $orange600;  
+    }
   }
   .contributor_name {
     line-height: 1.2;
--- a/app/assets/stylesheets/components/blank_slate.scss
+++ b/app/assets/stylesheets/components/blank_slate.scss
@@ -23,6 +23,12 @@
         height: 0;
         padding-bottom: 144%;
       }
+      &.alec {
+        background-image: image-url("illustrations/contributors/alec-talking.png");
+      }
+      &.jenya {
+        background-image: image-url("illustrations/contributors/jenya-talking.png");
+      }
     }
     .contributor_name.mobile_only, .contributor_title.mobile_only {
       display: none;
@@ -77,9 +83,4 @@
   background-position-y: $background_position-y;
   position: relative;
   top: -48px;
-  .left {
-    .blank_slate_illustration {
-      background-image: image-url("illustrations/contributors/alec-talking.png");
-    }
-  }
 }
--- a/app/assets/stylesheets/components/blank_slate.scss
+++ b/app/assets/stylesheets/components/blank_slate.scss
@@ -16,6 +16,7 @@
     .blank_slate_illustration {
       background-size: 100%;
       background-repeat: no-repeat;
+      background-position-y: bottom;
       height: 100%;
       min-width: 238px;
       @media only screen and (max-width: 540px) {
--- a/app/helpers/admin_form_builder.rb
+++ b/app/helpers/admin_form_builder.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+class AdminFormBuilder < ActionView::Helpers::FormBuilder
+  include ActionView::Helpers::FormTagHelper
+
+  attr_accessor :output_buffer
+
+  def field(attribute, options = {}, &block)
+    content = @template.capture(&block) if block_given?
+    options[:class] = options.key?(:class) ? Array(options[:class]) : []
+    options[:class] << 'field'
+    options[:class] << 'invalid' if @object.errors.include?(attribute)
+    tag.div(**options) do
+      concat(content) if content.present?
+    end
+  end
+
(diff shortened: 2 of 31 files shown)


Manfred:
Things that need some love:

* The public mass invite URL can be shorter
* The user form on the mass invite URL `app/views/mass_invites_users/new.html.erb`
* The mass invite archived warning page: `app/views/mass_invites_users/archived.html.erb`

spec/fixtures/mass_invites.yml
```diff
@@ -0,0 +1,28 @@
...
+  updated_at: <%= 1.year.ago %>
+cheese_eating_challenge:
+  name: Henri's Cheese Eating Challenge
+  token: OK22YhRGVbm6Vk4e
+  archived: false
+  users_count: 0
+  created_at: <%= 2.weeks.ago %>
+  updated_at: <%= 2.weeks.ago %>
+one_song:
+  name: One Song Per Hour Challenge
```
sudara, line 16: 🤣

app/helpers/admin_form_builder.rb
```diff
@@ -0,0 +1,30 @@
...
+class AdminFormBuilder < ActionView::Helpers::FormBuilder
+  include ActionView::Helpers::FormTagHelper
+
+  attr_accessor :output_buffer
+
+  def field(attribute, options = {}, &block)
+    content = @template.capture(&block) if block_given?
+    options[:class] = options.key?(:class) ? Array(options[:class]) : []
+    options[:class] << 'field'
+    options[:class] << 'invalid' if @object.errors.include?(attribute)
```
sudara, line 12: @ofsound In case you are wondering where the form classes in the admin are coming from...

app/assets/stylesheets/components/blank_slate.scss
```diff
@@ -23,6 +23,12 @@
         height: 0;
         padding-bottom: 144%;
       }
+      &.alec {
+        background-image: image-url("illustrations/contributors/alec-talking.png");
+      }
+      &.jenya {
+        background-image: image-url("illustrations/contributors/jenya-talking.png");
+      }
```
sudara, line 31: @ofsound Just a heads up that I'm going to add all the blank slate images here, so they are usable on all pages.

db/schema.rb
```diff
@@ -189,6 +189,23 @@
     t.index ["track_owner_id"], name: "index_listens_on_track_owner_id"
   end
 
+  create_table "mass_invite_signups", charset: "utf8mb4", force: :cascade do |t|
```
sudara, line 192: I have the `utf8mb4_unicode_ci` specified locally and everywhere in the schema but *not* in production, for some reason (perhaps it's default there?)

So I have some diff that looks like so:

```ruby 
create_table "mass_invites", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
```

This is just an informational notice, I don't really remember why things are this way, I will come back and look to see if it matters.

sudara: Context in https://github.com/sudara/alonetone/issues/732#issuecomment-532141963

Manfred: It matters when you sort text columns without specifying the collation or do full text search, otherwise it doesn't really matter. But I may be missing something.

Manfred: Past me is a lot smarter than current me.

config/routes.rb
```diff
@@ -104,6 +104,8 @@
   root :to => 'assets#latest'
 
   resources :users
+
+  get '/invite/:mass_invite_token', to: 'mass_invites_users#new', as: 'mass_invite'
```
Manfred, line 108: So what are you going to use for single-use invites 😄

sudara: Heh yeah, I know... I wanted to live with it for a day because this is the ideal URL...

sudara: Was thinking of using a route constraint to check token length or for a token prefix

db/schema.rb
Manfred: I'm still on MySQL 5.7 for a project to match production. We had the same inconsistency there, so it's either caused by my workstation setup or a difference in MySQL version.

sudara:
🥳

Issue #1094, opened 2021-01-22
Account password is reset and account locked on submit of Forget your Password form


ofsound:
Until the link in the **[alonetone.com] Change your password**  email aren't followed, the User will be booted from the site and receive a flash message of "It looks like your account is not active" if they try to log in with their current password.

PR #1097, opened 2021-01-24
Use find_by_login_method to configure Authlogic

Manfred:
Rails said:

```
DEPRECATION WARNING: find_by_login_method is deprecated in favor of record_selection_method, to avoid confusion with ActiveRecord's "Dynamic Finders". (https://guides.rubyonrails.org/v6.0/active_record_querying.html#dynamic-finders) For example, rubocop-rails is confused by the deprecated method. (https://github.com/rubocop-hq/rubocop-rails/blob/master/lib/rubocop/cop/rails/dynamic_find_by.rb) (called from <class:UserSession> at /Users/manfred/Code/alonetone/app/models/user_session.rb:2)
```

* Replaces with `find_by_login_method ` with `record_selection_method` when configuring the method in `UserSession`.
* Change `User.find_by_login_or_email` to use one query to find the user.
* Add some tests.

I tested this locally by logging in with login and email. I also made sure you can't login with a black username and valid password.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -178,8 +178,8 @@ class User < ApplicationRecord
     User.where(current_login_ip: user.current_login_ip).where('id != ?', user.id)
   end
 
-  def self.find_by_login_or_email(login)
-    User.find_by_login(login) || User.find_by_email(login)
+  def self.find_by_login_or_email(login_or_email)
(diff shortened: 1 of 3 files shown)


sudara:
Also super useful, didn't know about `record_selection_method`. Related: #1094 made me even more motivated to hand roll and own auth².

PR #1098, opened 2021-01-24
Use Ruby on Rails 6.1 way of iterating error messages

Manfred:
* Use the 6.1 way of iterating error message in the shared error partial.
* Use the shared error partial in views that had their own implementation.
* Remove dead error code form the assets controller.

This leaves one case that is in the Authlogic gem.


Manfred:
Visual changes are the order of the error messages.

sudara:
Thanks for this! These have been on my internal todo list. We are looking to unify the form error style shortly as well, so the consolidation is helpful.

Strangely, this is changing the order with which errors show up (hence the Percy check showed red), but its seems like the order didn't make sense to begin with.

PR #1103, opened 2021-02-04
Rewrite track reveal animation to GSAP, to fix window resizing bugs

### LINKED ISSUE #1105: Track chunk hitbox seems constrained
Basically, this area in red should open the track chunk when clicked.

Screenshot 2021-02-09 - Petere 🐺's Music | alonetone

Right now, only the top half opens the chunk.

Before, we had the problem with the bottom half opening the artist's page.

The intended interaction is that clicking/tapping anywhere on the track chunk that isn't the play button or the artist name will open the chunk.

ofsound:
Fixes #1105

## TODO 

* [ ]  **_check that the cache is happy_**
* [ ] Fix failing tests on CI

**_Resizing the browser window both larger and smaller after a track reveal had been opened and closed exposed an issue with making only a one time height calculation of the reveal. Partial bits of the track_reveal could be seen peaking out when they should have remained hidden_**

This solution will hide the track reveal completely right after its closing animation.

Upon re-opening it will re-calculate a position exactly off screen to start its animation back to margin-top: 0px

I added an extra class to let the CSS know about the special asset case of a track with the favorited by section.

The 'open' class is still used to set other CSS properties, but is no longer a trigger for a CSS transition.

I slightly re-wrote the currentlyOpen section of the JS, before there was a conditional in the openDetails which seemed unnecessary.

I attempted to have a single GSAP timeline with labels in the initialize(), but it wasn't calculating the offsetHeight correctly, so I kept a single GSAP timeline to avoid multiple tween directions, but have the open and close methods clearing that timeline and applying their transitions.

The homepage spec is breaking, but it doesn't seem too serious. Maybe elements are just offscreen in a slightly different way or order. We'll see....

* [ ] All tests green


app/javascript/controllers/normal_playback_controller.js
```diff
@@ -38,6 +43,7 @@ export default class extends PlaybackController {
       // if another track details is open, close it
       if (currentlyOpen) {
         currentlyOpen.closeDetails()
+        currentlyOpen = undefined
```
sudara, line 46: Nice catch with this. We should prob go one step further and move the assignment into `closeDetails()` so that it conceptually parallels the assignment in `openDetails()`. Also, how do you feel about `false` vs. undefined?....

sudara (approved):
Nice work! Cool to have this in gsap, feels snappy. I did some lite QA at different browser sizes and on mobile to see how it feels re: opening chunks at the extreme top/bottom, looks good.

sudara:
Looks like there's a css merge conflict....

sudara:
Dug deep here. I'm not 100% sure why the tests are only failing on the branch actually. They fail locally when the elements are out of view. Then capybara has problems being told to click on something unless it's explicitly passed `visible: :all` which will let it click on something below the fold or otherwise hidden.

Added some details about how to run those feature tests locally so you can try it out next time. That actually had broken though, so I got that happy again, cleaned up the config, etc. 😰

sudara:
Hmm, tests still fail on CI, but they are passing locally. :(

app/javascript/controllers/normal_playback_controller.js
ofsound: @sudara I thought about that distinction and chose undefined because the assignment is an object if it exists, rather than a true. But I'm not too invested either way.

sudara:
This particular error doesn't seem to show up until yesterday. I believe it might be related to me manually setting the screen size that CI is running at.  I do get this problem locally with the seeds. Ideally we'd bypass fixing it because it'll move to a new paradigm after #1087 

```
Failure/Error: find('.play_button_container a').click # pause
```

sudara:
Closing in favor of #1108 — nice work on the js stuff!

PR #1107, opened 2021-02-23
Include MorphSVGPlugin in percy specs

sudara:


sudara:
Not 100% sure this functioned as intended, but what they hey, going to merge anyway and see what happens down the line.

PR #1108, opened 2021-02-23
Gsap animation redux

### LINKED ISSUE #1105: Track chunk hitbox seems constrained

sudara:
Continues #1107 
Fixes #1105

This fixes various issues on CI related to javascript, capybara and percy:

* Coverts track chunk css animation to js gsap animation
* Adds a test to fail CI on javascript error
* Adds more verbose output for javascript warnings and errors
* Fixes webpacker failing to output when asset compilation fails
* Fixes webpacker cache not persisting across CI runs
* Documents how to run feature specs locally in CONTRIBUTING
* Updates WebMock config to allow for other capybara drivers

Doesn't handle:

* When a manual screen size is specified, we get an additional playlist spec failing, that same problem as locally with the playlist track evaporating.
* Currently manually ignoring js errors about CORS, uncertain whether they are occurring on CI or just locally

```
# SEVERE 2021-02-24 17:55:00 +0100: https://cdn.alonetone.com/fonts/Alright-v2-Normal-Bold-latin1-tnum.woff2 - Failed to load resource: net::ERR_FAILED
```


sudara:
@ofsound This test you did (that I just repeated) of setting `this.detailsTarget.style.marginTop = 0;` was a good idea.

We know js is working, as we see the open class get added. 

However, we get the same result (track doesn't open) whether using gsap or raw js, chrome or FF.

I will try another test to double check gsap is happy.

sudara:
To be clear: it failed to open the track chunk in CI, succeeded locally. It's possible there's a problem with offsetHeight there? 

However, the reason the specs are failing is because every track chunk is always technically visible (css-wise). The specs rely on waiting for elements to become visible to "prove" that actions have occurred. So we'll need to maintain what is happening on master, which is that that the detailsTarget (reveal) is being toggled between display none and display block.

sudara:
Hmm, guess i was wrong that the tests are failing because things weren't hidden. Maybe they are failing because the offset height isn't present on CI somehow?

sudara:
Wow, even with a hardcoded margin, it seems to be failing. I'm at a loss again....

sudara:
@ofsound All done. 😭 Thanks for all the help debugging! 🥇

PR #1113, opened 2021-03-02
Fixes for .flash padding, track description padding, footer collapsing

### LINKED ISSUE #1104: Footer has strange inbetween state
Around 957px wide on Safari

Screenshot 2021-02-09 - alonetone

### LINKED ISSUE #1109: Too much top padding when no description in track chunk
Screenshot 2021-02-24 - Latest Music | alonetone

The uploaded date should be vertically aligned with where the description would be. Prob makes most sense to add a ruby class when no description present?

ofsound:
Fixes #1109
Fixes #1104

*Deploying requires a cache clear*

Fixes an issue I've noticed on iPad where a flash message is the exact width of the screen, revealing the lack of padding.

Seems that extra padding ( actually a min-height ) of the user_description was also preserving the bottom alignment of the Track details and Listen and Favorite section within an accordion track.

I could maybe separate that section from the top down lay outing, and have it stuck to the bottom of that container rather than preserving a fixed distance below the date, or it could float a bit higher as it appears in the 2nd screenshot.

Screen Shot 2021-03-02 at 4 54 51 PM

Screen Shot 2021-03-02 at 4 54 45 PM

* [ ] All tests green

### DIFF
--- a/app/assets/stylesheets/components/notifications.scss
+++ b/app/assets/stylesheets/components/notifications.scss
@@ -14,7 +14,7 @@
   line-height: 1.2;
   position: relative;
   z-index: 101;
-  padding: 15px 0;
+  padding: 15px 15px;
   text-align: center;
   color: $flash-text;
   background-color: $accent;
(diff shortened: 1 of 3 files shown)


sudara:
> I could maybe separate that section from the top down lay outing, and have it stuck to the bottom of that container rather than preserving a fixed distance below the date, or it could float a bit higher as it appears in the 2nd screenshot.

Yeah, I think that's a good idea if it's not too much of a pain. We should preserve the "View Track Detail" and the heart/play icon spacing with regards to the bottom of the container...

PR #1133, opened 2021-06-28
Pull in latest topics from discourse

sudara:
[insert @ofsound magic here]

We can pull in the latest topics and style with discourse's embedded css: https://forum.alonetone.com/admin/customize/themes/1/common/embedded_scss/edit

Supporting both themes might be awkward? Not sure it's going to be possible to get that info through the iframe without some js...

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -10,6 +10,8 @@
   <%= stylesheet_link_tag 'application', disabled: dark_theme? %>
   <%= stylesheet_link_tag 'dark_theme', disabled: light_theme? %>
   <%= javascript_pack_tag "application", 'data-turbolinks-track': 'reload', defer: true  %>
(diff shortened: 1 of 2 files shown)


sudara:
> just seems like we’d be fighting every single bit of the iFrame, from fonts, to dynamic media queries, to white/theme dark theme

If we don't go the iframe way, options are:

1. Parse json client-side every request, ala
https://meta.discourse.org/t/embed-latest-post-teasers-into-your-website/82388

2. Parse json in rails, create and cache a partial 

3. Setup a ruby cron to create json from the html, stick it in memcached with a key for Rails to read later

sudara:
Option 1 could be accomplished by issuing an ajax request on a stimulus controller `initialize`, storing the result in a member variable,  moving the footer template to js and rendering the template in `connect`.

PR #1146, opened 2021-10-02
add mysql install step to contrib doc

jlmitch5:
this was needed to get past a bundle install error

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,6 +11,7 @@ You will need to first install [Homebrew](https://brew.sh).
 - `brew install libsndfile lame` (required for id3 tags and waveforms)
 - `brew install vips` (required for processing images)
 - `brew install yarn` (required for building assets)
+- `brew install mysql` (required for local database)
 - Install gems
 `bundle install`
 - Install front-end dependencies


jlmitch5:
```
- Create and seed database
`rake db:setup`
```

I'm assuming the above rake command is ran as part of `rake setup`...this is just a subset if the only thing you need to do is setup the db?  It doesn't run for me because:

`Database 'alonetone_development' already exists` which is why I'm thinking that.

if so, I think this step could probably be removed from the contrib.

jlmitch5:
dev env setup success!

Screen Shot 2021-10-01 at 10 20 20 PM

sudara:
> if so, I think this step could probably be removed from the contrib.

You are right `rake db:setup` is redundant!

CONTRIBUTING.md
```diff
@@ -36,7 +45,7 @@ The frontend code can take advantage of `MorphSVGPlugin` for more fluid SVG anim
 
 ## Logging in
 
-After the bootstrap data is loaded, you can login using the test account. Username is "admin" and password is "testing123"
+After the bootstrap data is loaded, you can login using the test account. Username is "owner" and password is "testing123"
```
sudara, line 48: Nice catch. We could go a step further and say that the password for any of the seed users (including owner) is "testing123".

CONTRIBUTING.md
```diff
@@ -11,6 +11,15 @@ You will need to first install [Homebrew](https://brew.sh).
...
 - `brew install yarn` (required for building assets)
+- `brew install mysql` (required for local database)
+  - before running rake to generate the data, you'll also need to start the mysql server, as well as have the root user with password 'root' and a currently logged in user with no password.
+  - you can find currently logged in user by doing: `stat -f '%Su' /dev/console` will refer to this as currentUser
+  - `mysql.server start`
+  - `mysql -u root -h localhost` to open root mysql prompt.  In the prompt type the following lines, replacing currentUser with what you figured it out to be above
+  - `SET PASSWORD FOR root@localhost='root';`
+  - `CREATE USER 'currentUser'@'localhost' IDENTIFIED BY '';`
+  - `GRANT ALL PRIVILEGES ON * . * TO 'currentUser'@'localhost';`
+
```
sudara, line 22: Good idea to add info on mysql.The instructions after `myqsl.server start`  shouldn't be  necessary, as mysql ships with a root user and no password, which matches the setup that alonetone expects. So I believe if you re-did the setup from scratch and didn't want to use a custom user, it should work out of the box.

jlmitch5: Cool!  I do think the root password may need to be updated.  It looks like it was changed to 'root' because this is what github actions expects, and that is different than the brew default of user 'root' not having password.  I removed the other lines though.  Happy to change if I'm not understanding something.

CONTRIBUTING.md
jlmitch5: > Nice catch. We could go a step further and say that the password for any of the seed users (including owner) is "testing123".

👍  added those.

jlmitch5:
> > if so, I think this step could probably be removed from the contrib.
> 
> You are right `rake db:setup` is redundant!

removed

jlmitch5:
thanks for the feedback, I added those.

PR #1147, opened 2021-10-02
Update yarn lockfile and add node dep to contrib doc

jlmitch5:
I was having trouble installing node deps at a single version of node (10 needed to get pass fsevent build error, 12 needed to satisfy percy version requirement).  I bumped the lock file, which bumped minor version on all the deps based on the package.json ^, and it build fine on 12 after that.

I update the contrib doc to indicate what version of node to use.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,7 @@
 ### Requirements
 
 Ruby 2.7.x
-
+Node 12.22.x
 ### MacOS
 
 You will need to first install [Homebrew](https://brew.sh).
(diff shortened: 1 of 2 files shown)


sudara:
Do you happen to still have the backtrace/error? Mainly curious.

I think production is 12.18, so we'll have to coordinate a node bump there. Would it make sense to bump to a higher version while we are at it?

Related: I'm currently updating all js deps over in #1087.

jlmitch5:
> Do you happen to still have the backtrace/error? Mainly curious.

I don't believe I do...it was after install on the node-gyp build step for fsevents.  This has happened to me before, which is why I guessed it was node version related.  So I just tried different `nvm use $versions` until I got it to work.

> I think production is 12.18, so we'll have to coordinate a node bump there. Would it make sense to bump to a higher version while we are at it?

Yep probably worth it to try 14 or 16 (after doing `rm -rf node_modules` and then rerunning yarn to make sure all depedency building works fine)

https://nodejs.org/en/about/releases/

> Related: I'm currently updating all js deps over in #1087.

Ah gotcha 👍 

In that case, I'll close this.  I think it would be good to call out what version of node you expect to run with.

Issue #1196, opened 2022-04-11
Allow users to set a license for their tracks


Manfred:
Sudara said, "I would like to add a license option to alonetone."

The default license for all tracks (assets in Alonetone) should be "All rights reserved", meaning they retain all the rights granted to then according to the laws in affect while uploading the audio.

Alternatively users can choose to change the license for a track (asset in Alonetone) to a Creating Common license by selecting it from a list of options. We could support other licenses in the future if the need arises.

We can also allow users to change the license for an entire album (playlist in Alonetone) by setting the license on the album. This will not be persisted in the database, but is forwarded to the tracks (assets).

Once of the considerations is that Creating Commons licenses have a version and that may change in the future. We will store the version select by the user, but should also be able to compare it to other versions.

As a follow-up to this issue we should write an issue about browsing tracks with a specific license.

PR #1197, opened 2022-04-11
Allow users to set a license for each track separately

Manfred:
* Adds a column to assets to store a license code
* Adds a proxy class for license to format license related labels and icons
* Adds a license option to the asset form
* Support for dark and white themes

References #1196

### DIFF
--- a/app/assets/stylesheets/components/single_and_playlists_player/track_content.scss
+++ b/app/assets/stylesheets/components/single_and_playlists_player/track_content.scss
@@ -215,6 +215,22 @@
           }
         }
       }
+      .license {
+        a {
+          width: 106px;
+          margin-right: 8px;
+          svg {
+            height: 24px;
+            margin-left: 6px;
+            margin-top: 1px;
+            top: 0;
+            fill: $track-content-license-text;
+            circle {
+              fill: none;
+            }
+          }
+        }
+      }
     }
   }
   .track_edit {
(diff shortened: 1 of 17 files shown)


Manfred:
Screen Shot 2022-04-11 at 22 08 46
Screen Shot 2022-04-11 at 22 09 01
Screen Shot 2022-04-11 at 22 09 10
Screen Shot 2022-04-11 at 22 09 23

Manfred:
I also considered using the default Unicode symbols instead of SVG, but I'm pretty sure those aren't supported on a lot of devices. https://creativecommons.org/2020/03/18/the-unicode-standard-now-includes-cc-license-symbols/

sudara:
This all looks great! 

Oooh, that's cool about unicode CC symbols. SVG is perfect though, more pragmatic for now.

sudara:
Got @smoofles over and we're taking a look at cleaning up the "show track" page a bit and making sure this all looks good on mobile, etc! Thanks again for the feature @Manfred!

sudara:
Added #1199 and #1200 to summarize design cleanup to assets#show and assets#edit to enable

@bencarlos — probably easiest if those changes just happen on this branch. Ping me if you think I need to move some rails logic around for fave and edit, otherwise i think it should be pretty straightforward!

sudara:
@Manfred We tried to include CC icons in a couple places, but in the end they seemed to mainly add visual bulk/noise, so I think they didn't make the cut... 😢 

image

PR #1239, opened 2024-02-11
[ISS-1111] Add reserved words check to profile registration

### LINKED ISSUE #1111: Reserved Words
In order to keep nice urls (alonetone.com/sudara) we'd need to create a deny list of reserved words, including anything that might be in routes.rb.

em-wilson:
Resolves [Issue-1111: Reserved words](https://github.com/sudara/alonetone/issues/1111) - In order to keep nice urls (alonetone.com/sudara) we'd need to create a deny list of reserved words, including anything that might be in routes.rb.

Proposed Changes
===============
1. Adds a helper to check if a username is valid
* Names must not be in use by an account (it's ok if you're already the person using the name)
* Names must not be in use by an account request
* Names must not be part of a DENY list
* Names must not be part of a route, or partial route, provided by the rails application
image

2. Performs a username check every time someone has an opportunity to change their user name
* When signing up
* When editing their profile
* No difference in user message between valid & taken names is on purpose; I like obfuscating the reason to discourage path stuffing attacks
image

3. Adds a CRUD screen to add to the deny word list
* Words allow REGEX; it's possible to differentiate between "tomshitfactory" and "expletiveshit" for example
image

### DIFF
--- a/app/assets/stylesheets/components/admin.scss
+++ b/app/assets/stylesheets/components/admin.scss
@@ -273,7 +273,9 @@
       padding: 0.375em 0.5em;
     }
   }
-  form.new_mass_invite {
-    padding: 0 1em;
+  form {
+    &.new_mass_invite, &.new_reserved_word {
+      padding: 0 1em;
+    }
   }
 }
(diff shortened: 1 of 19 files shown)


sudara:
Thanks for submitting this! I took a break from maintaining for a bit, but I'm back and incorporated your work into #1247 — One thing I'm curious about is where the [ISS-1111] prefix came from! Anyway, closing this for now, but your commit is kept on the new PR!

